PR #178, opened 2017-10-02
Cleanup internationalization, removed old custom posix

tintou:


src/util/system.vala
```diff
@@ -18,7 +18,7 @@
 */
 
 int number_of_processors () {
-    int n = (int) ExtendedPosix.sysconf (ExtendedPosix.ConfName._SC_NPROCESSORS_ONLN);
+    int n = (int) Posix.sysconf (Linux._SC_NPROCESSORS_CONF);
```
davidmhewitt: So I'm not sure where this `number_of_processors` method is used, but if it's used for determining the number of threads to use for something, we should probably keep it as `ONLN` instead of `CONF`. In case the operating system has disabled some processors, we should only be threading across the online ones.

davidmhewitt:
Also, the `ExtendedPosix.vapi` file could be removed now?

tintou:
Ah sorry, this is what I intended to do :stuck_out_tongue:

PR #182, opened 2017-10-12
HiDPI thumbnail support

### LINKED ISSUE #133: Provide HiDPI thumbnails
Currently, the thumbnails are lowDPI at every scale factor…

Launchpad Details: [#LP1661783](https://bugs.launchpad.net/bugs/1661783) Corentin Noël - 2017-02-04 00:41:12 +0000

davidmhewitt:
Fixes #133 

:tada:

### DIFF
--- a/src/Checkerboard/CheckerboardItem.vala
+++ b/src/Checkerboard/CheckerboardItem.vala
@@ -89,21 +89,6 @@ public abstract class CheckerboardItem : ThumbnailView {
         }
     }
 
-    private int _scale_factor = 1;
-    private int scale_factor {
-        get {
-            return _scale_factor;
-        }
-        set {
-            if (_scale_factor != value) {
-                _scale_factor = value;
-                ThumbnailCache.scale_factor = _scale_factor;
-                recalc_size ("set_scale_factor");
-                notify_view_altered ();
-            }
-        }
-    }
-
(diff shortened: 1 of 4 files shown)


danrabbit:
I can confirm that this makes thumbnails much sharper on HiDPI :)

davidmhewitt:
This should be *much* better now. Ready for review!

danrabbit:
On first run I got this once and then it never happened again:

```
ERROR:/home/daniel/Projects/photos/src/ThumbnailCache.vala:493:_thumbnail_cache_import_thumbnail: assertion failed: (Dimensions.for_pixbuf (scaled).approx_scaled (size.get_scale ()))
Aborted (core dumped)
```

davidmhewitt:
@danrabbit The only thing I can think that may have caused that is the fact that you tested this before I pushed another commit that fixed a couple of things.

I've just tested from clean, by deleting everything out of `~/.cache/pantheon-photos/thumbs/` and then running this branch and also by running master first, letting it generate some thumbnails in that folder, then running this branch again and experienced no issues. This simulates both the upgrade and clean install processes.

Would you be able to do the same and confirm that it's not one particular image that's causing the issue?

danrabbit:
I cleared the cache and it seems to be working as expected. No crash :)

Issue #62, opened 2017-04-18
Raw Developer says "Shotwell", not "Photots"


elementaryBot:
In Pantheon Photos settings, there is an option for "Raw Developer"
The option should say "Photos", but has the old name ("Shotwell").

This is confusing for new users, as there is now nothing named Shotwell in Elementary.

ProblemType: Bug
DistroRelease: elementary OS 0.3
Package: elementary-desktop 1.349+393~ubuntu0.3.1 [origin: LP-PPA-elementary-os-daily]
ProcVersionSignature: Ubuntu 3.13.0-34.60-generic 3.13.11.4
Uname: Linux 3.13.0-34-generic x86_64
ApportVersion: 2.14.1-0ubuntu3.3
Architecture: amd64
CrashDB: elementary_meta
CurrentDesktop: Pantheon
Date: Wed Aug 20 10:56:35 2014
InstallationDate: Installed on 2014-08-11 (8 days ago)
InstallationMedia: elementary OS 0.3 "Freya" - Daily amd64 (20140810)
SourcePackage: elementary-meta
SuspiciousXErrors:
 
ThirdParty: True
UpgradeStatus: No upgrade log present (probably fresh install)

Launchpad Details: [#LP1359114](https://bugs.launchpad.net/bugs/1359114) Lars-Eivind Bjørtvedt - 2014-08-20 09:01:44 +0000

PR #186, opened 2017-10-15
Greatly reduces mentioning of "Shotwell"

### LINKED ISSUE #96: we need to change every instance where "Shotwell" appears to "Photos" to maintain consistency.
We don't use the "Shotwell" name anymore, although apareça constantemente. So we need to change that to "Photos" instead. 

e.g. 

"You have already logged in and out of a Google service during this Shotwell session." 

Should be:

"You have already logged in and out of a Google service during this Photos session."

Launchpad Details: [#LP1421628](https://bugs.launchpad.net/bugs/1421628) Davi A. - 2015-02-13 12:30:32 +0000

snowparrot:
Fixes #96 #62 

This PR replace Shotwell with Photos in many places, mostly in comments and strings. 
I didnt refactor and I didnt change ressources.vala, because it may be more important in code and PublishingInterface.vala, because by logic SPIT needs to change into PPIT :P


codygarver:
Potential reviewers: please don't merge this until I've made a new stable release of Photos, since this is gonna break some translations.

tintou:
I'm in favor of a complete removal of any personification of the application as much as possible
and in some place things can be reformulate like:
`Move the photos to the Photos trash` → `Move photos to the photos trash` (it's not the trash of the application but the trash dedicated to photos items)

tintou:
(talking about user-facing string naturally)

danrabbit:
I'm going to merge this just because it's better than master. But I agree that these strings could probably be a lot better and they don't have to self-reference

PR #194, opened 2017-11-04
Use granite debugging/logging instead of custom class

davidmhewitt:
It was a bit of a pain to try and remember which environment variable you had to set to get debug info out of photos. This brings it more in line with other elementary flags where you just pass a `--debug` flag and get nice coloured Granite debug messages in the console.


src/main.vala
```diff
@@ -315,6 +316,11 @@ public OptionEntry[] get_options () {
                           };
     entries += version;
 
+    OptionEntry debug_option = { "debug", 'D', 0, OptionArg.NONE, &debug_enabled,
+                            _ ("Show extra debugging output"), null
+                          };
```
jeremypw: Indenting does not match other OptionEntry lines.

jeremypw:
I think it would need more extensive changes as the option entries are defined in main?? Also the option -d is already used.  I notice that using the --help option does not show --help-granite (compare e.g. AppCenter) but I am not sure why.

davidmhewitt:
@donadigo Jeremy is right. Photos does its own parsing of command line arguments and they never get passed to Granite.Application.

It would require some quite extensive rework to enable Granite to add the `--debug` flag itself, and there would be quite a high chance of causing a regression due to the strange way command line parsing is done in Photos. So, I'd rather not in this case.

Formatting fixed, however.

src/main.vala
```diff
@@ -315,6 +316,11 @@ public OptionEntry[] get_options () {
                           };
     entries += version;
 
+    OptionEntry debug_option = { "debug", 'D', 0, OptionArg.NONE, &debug_enabled,
```
donadigo: Shouldn't `ref debug_enabled` be used here instead of this C syntax?

donadigo:
@davidmhewitt ah that makes sense.

davidmhewitt:
@donadigo I could use `ref` there but then it would be inconsistent with the rest of the code for the command line params. I'd rather leave that one like that or change all of them. Your call! :)

PR #195, opened 2017-11-04
Increase thumbnail cache size limit

davidmhewitt:
Since the addition of the ability to generate/display HiDPI thumbnails, these larger thumbnails were not able to fit in the cache. Increase the size limit of individual thumbnails by x4 (4 times more pixels!).

Also increase size limit of all cached thumbnails combined to 100MB across both large and small thumbnails to prevent as much thrashing of the disk when scrolling back to previously loaded thumbnails.


jeremypw:
Seems reasonable but needs testing with HiDPI screen and thumbnails.

PR #202, opened 2017-11-07
Make +/- zoom buttons incremental

### LINKED ISSUE #177: Revise behavior of zoom in / out icons
When clicking either the plus or minus icons in the lower-right to zoom in or out, the photo is either zoomed in as far as it will go, or zoomed all the way out of.  This behavior strongly deviates from incrementally zooming in or out, which is what the plus and minus icons imply will happen.

It may require crafting two new icons to handle zooming all the way in or out, but it makes much more sense for the plus and minus icons to zoom in steps, with fine adjustments made on the slider.

davidmhewitt:
Fixes #177 

Uses a couple of timeouts to allow for nice press and hold behavior on the buttons. Reads GTK settings to get the timeout values.


src/MediaPage.vala
```diff
@@ -101,16 +111,49 @@ public abstract class MediaPage : CheckerboardPage {
...
+                increase_step ();
+            } else {
+                decrease_step ();
+            }
+
+            return true;
+        }
+
+        private bool on_zoom_button_pressed (Gdk.EventButton event, Gtk.EventBox sender) {
+            clear_zoom_timeouts ();
```
jeremypw: Probably not required as timeouts cancelled on button release.

src/PhotoPage.vala
```diff
@@ -535,12 +542,13 @@ public abstract class EditingHostPage : SinglePhotoPage {
...
             Gtk.Image zoom_out = new Gtk.Image.from_icon_name (Resources.ICON_ZOOM_OUT, Gtk.IconSize.MENU);
-            Gtk.EventBox zoom_out_box = new Gtk.EventBox ();
+            zoom_out_box = new Gtk.EventBox ();
             zoom_out_box.set_above_child (true);
             zoom_out_box.set_visible_window (false);
             zoom_out_box.add (zoom_out);
 
-            zoom_out_box.button_press_event.connect (on_zoom_out_pressed);
+            zoom_out_box.button_press_event.connect ((event) => on_zoom_button_pressed (event, zoom_out_box));
+            zoom_out_box.button_release_event.connect ((event) => on_zoom_button_released (event, zoom_out_box));
```
jeremypw: You do not need to use a closure:
```
zoom_out_box.button_press_event.connect (zoom_button_pressed);
zoom_in_box.button_press_event.connect (zoom_button_pressed);
```
Will pass the event source to the handler. You will need to change the handler signature to:
```
private bool on_zoom_button_pressed (Gtk.Widget sender, Gdk.EventButton event) {}
```
Then cast sender to Eventbox.

src/PhotoPage.vala
```diff
@@ -654,16 +663,49 @@ public abstract class EditingHostPage : SinglePhotoPage {
...
 
-    private bool on_zoom_in_pressed (Gdk.EventButton event) {
-        snap_zoom_to_max ();
+    private bool on_zoom_button_released (Gdk.EventButton event, Gtk.EventBox sender) {
+        clear_zoom_timeouts ();
+
         return true;
     }
 
+    private void clear_zoom_timeouts () {
```
jeremypw: Should really call this on leave-notify event in case button has not been released. (Perhaps unlikely but possible)

davidmhewitt:
Thank you for the review, @jeremypw . The code is much nicer and less repetitive now as a result of the changes you suggested.

src/Widgets/ZoomSliderAssembly.vala
```diff
@@ -0,0 +1,150 @@
...
+        zoom_in_box.button_press_event.connect (on_zoom_button_pressed);
+        zoom_in_box.button_release_event.connect (on_zoom_button_released);
+        zoom_in_box.leave_notify_event.connect (() => {
+            clear_zoom_timeouts ();
+
+            return false;
+        });
+
+        add (zoom_in_box);
+    }
```
jeremypw: Nearly all of this can be put in the construct clause by creating a slider with null adjustment parameter and then just applying the min, max, step and initial_val in new ():

```

    public ZoomSliderAssembly (double min, double max, double step, double initial_val) {
        slider.set_range (min, max);
        slider.set_increments (step, step);
        zoom_value = initial_val;
    }
```

davidmhewitt:
@jeremypw Thanks! Makes sense, I've made those changes.

PR #204, opened 2017-11-09
Port library prefs from ui file to vala

danrabbit:


davidmhewitt:
Not sure on merging yet since this is presumably going to break translations and Photos hasn't had a release in a while. So I'll leave it up to whoever. But it works good and looks good!

danrabbit:
Sounds like a question for @codygarver

PR #219, opened 2017-11-29
Remove editing GSettings

davidmhewitt:
I was porting these settings to native GSettings as I've been doing with the others. When it came to testing them, I noticed that they weren't really being used.

I think the original intention of these was that once you'd right clicked an image and said "Open With", it stored that application in GSettings. Then it looked like at some point in the past, it was possible to hold super and double click an image in your library and it would open in that "default" editor stored in GSettings.

That's been broken in master for who knows how long and seems like a bit of a hidden/confusing feature anyway. So I propose removing the GSettings and associated code altogether as per this branch.


danrabbit:
tenor

PR #222, opened 2017-12-01
Simplify app options code

Thalexander:
Have a const to store the app options rather than using a function to generate it, put the variables in the correct order to match that of the options and use lowercase letters for the short names.


src/main.vala
```diff
@@ -284,48 +284,20 @@ void editing_exec (string filename) {
...
-    return entries;
-}
+    string data_dir = null;
+    bool no_runtime_monitoring = false;
+    bool no_startup_progress = false;
+    bool show_version = false;
+    bool debug_enabled = false;
+
+    public const OptionEntry[] app_options = {
+        { "datadir", 'd', 0, OptionArg.FILENAME, out data_dir, N_("Path to Shotwell's private data"), N_("DIRECTORY")},
```
danrabbit: While we're here, let's change this to "Photos'" instead of "Shotwell's" :)

src/main.vala
```diff
@@ -284,48 +284,20 @@ void editing_exec (string filename) {
...
+    bool no_startup_progress = false;
+    bool show_version = false;
+    bool debug_enabled = false;
+
+    public const OptionEntry[] app_options = {
+        { "datadir", 'd', 0, OptionArg.FILENAME, out data_dir, N_("Path to Photos' private data"), N_("DIRECTORY")},
+        { "no-runtime-monitoring", 0, 0, OptionArg.NONE, out no_runtime_monitoring, N_("Do not monitor library directory at runtime for changes"), null},
+        { "no-startup-progress", 0, 0, OptionArg.NONE, out no_startup_progress, N_("Don't display startup progress meter"), null},
+        { "version", 'v', 0, OptionArg.NONE, out show_version, N_("Show the application's version"), null},
+        { "debug", 'd', 0, OptionArg.NONE, out debug_enabled, N_("Show extra debugging output"), null},
```
davidmhewitt: I think the `short_name` needs to be put back to upper case here, or just removed entirely. It's currently conflicting with `datadir`.

PR #228, opened 2017-12-06
Don't bind null pointers to database queries

### LINKED ISSUE #193: database bind text error for each file on importing new photos
bildschirmfoto von 2017-11-02 19 29 35

davidmhewitt:
Fixes #193 

The SQLite library was throwing assertions about trying to bind null pointers to strings in database queries.

Should be able to test this by running from the console and importing a new photo. The current master will probably spam the console with errors.

### DIFF
--- a/src/db/DatabaseTable.vala
+++ b/src/db/DatabaseTable.vala
@@ -180,7 +180,11 @@ public abstract class DatabaseTable {
         assert (res == Sqlite.OK);
     }
 
-    protected static void bind_text (Sqlite.Statement stmt, int column, string data) {
+    protected static void bind_text (Sqlite.Statement stmt, int column, string? data) {
+        if (data == null) {
+            bind_null (stmt, column);
+        }
+
         var res = stmt.bind_text (column, data);
         assert (res == Sqlite.OK);
     }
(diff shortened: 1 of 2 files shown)


src/photos/PhotoMetadata.vala
```diff
@@ -868,8 +868,11 @@ public class PhotoMetadata : MediaMetadata {
         }
     }
 
-    public override string? get_comment () {
-        return get_string_interpreted ("Exif.Photo.UserComment", PrepareInputTextOptions.DEFAULT & ~PrepareInputTextOptions.STRIP_CRLF);
+    public override string get_comment () {
+        var comment = get_string_interpreted ("Exif.Photo.UserComment",
+                                              PrepareInputTextOptions.DEFAULT & ~PrepareInputTextOptions.STRIP_CRLF);
+
+        return comment != null ? comment : "";
```
jeremypw: This can be written `return comment ?? "" `

davidmhewitt:
@jeremypw Well, I learned something new today, thanks!

Seems like quite a nice operator to remember!

davidmhewitt:
Please don't merge this yet, I think I've had a better idea on how to do this.

davidmhewitt:
This is now ready for review. When the `bind_text` method was previously being passed null pointers and throwing errors, it was actually binding `NULL` to those values in the database, so it makes sense to do that properly (i.e. without throwing errors) rather than modifying the way the data is stored by storing empty string as that may have some impact on queries etc...

PR #239, opened 2017-12-10
PngSupport.vala: Remove workaround for gdk.pixbuf scale

danrabbit:
Removing this workaround seems to fix the crash on Artful with Vala 0.39


danrabbit:
This is garbage

PR #244, opened 2017-12-12
Rewrite slideshow settings dialog

davidmhewitt:
Essentially, I've replaced the slideshow settings dialog which looked like this and used to be opened by a cog icon on the slideshow toolbar:
screenshot from 2017-12-09 01 04 33 cropped

With this:
screenshot from 2017-12-18 00 28 16

There are a few UX questions that come to mind:

1. Is it OK to replace the two sliders for transition duration and delay between photos with a single speed slider, where right is faster and left is slower? I thought we might make the transition effect duration a function of the speed, so the faster the slideshow, the snappier the effects.
3. We also maybe need a different icon on the right for the toggle that denotes whether image titles should be shown or not? I kind of chose something that vaguely denoted some text, and all of the controls are tooltipped.
4. Any other thoughts?


danrabbit:
Nice! I like this. Much simpler and makes that toolbar much more useful. I pushed a commit to make that combobox a little more combobox-y. Maybe we should use a Gtk.MenuButton or a Gtk.ToggleButton there so that it stays depressed while the popover is open.

I agree that a tortoise and hare set would be better than + and -. Until someone can draw those up, maybe we can just use markers with "Faster" and "Slower" text labels?

One speed slider is probably fine.

I'm not sure what to do about the "Show Titles" selection, but the button is a little weird.

I think the slider should probably be at the end (before un-fullscreen)

davidmhewitt:
@danrabbit I've made those changes, should be ready for review.

PR #254, opened 2017-12-20
CheckerboardLayout: render the background colour explicitly

davidmhewitt:
This should only do anything on Juno. Most of the pages there either have a bright white or black background instead of the proper dark grey as seen on Loki.

Seems there were some changes in the newer versions of GTK that left us with the wrong coloured background on most of the pages. Take the background colour from the stylesheet and paint that on first.


davidmhewitt:
This may now need some changes in the stylesheet as we're shipping the same classes there. I don't remember if the plan was to remove the classes from here or the main stylesheet in the future.

Pinging @danrabbit for review since he understands this stylesheet stuff better than I do.

danrabbit:
Hey, this branch seems to work here :)

There is a checkerboard style class in the stylesheet because it's used in other places (like in screenshot tool), but we didn't end up creating a Granite style constant for it yet. So I would ignore it and pretend it doesn't exist and do what needs to be done in the embedded stylesheet for now

PR #255, opened 2017-12-20
RDNN

### LINKED ISSUE #123: use the right reversed elementary Scheme
We do not own pantheon.org, so our GSettings living under org.pantheon does not follow RDNN (https://en.wikipedia.org/wiki/Reverse_domain_name_notation).

In Loki+1, we should switch to using io.elementary for shell/OS components, and com.github.elementary for first-party apps.

Things to update:
 * Application id
 * Desktop filename
 * AppStream filename and id
 * GSettings path and id
 * DBus interfaces

Launchpad Details: [#LP1594598](https://bugs.launchpad.net/bugs/1594598) Cassidy James Blaede - 2016-06-20 22:26:20 +0000

mdh34:
RDNN files - fixes #123


danrabbit:
Hey thanks for your branch!

I'm marking this as `blocked` for now because changing the name of the .desktop file will remove it from the dock on update and changing gsettings schema means users settings will be effectively reset to defaults.

If you wanted to submit branches with some of the other changes that won't effect users (like renaming the binary and the gresource paths and even the appdata.xml file), we could merge those changes in separately :)

mdh34:
Ah ok, will do

danrabbit:
I'm going to close this because it has a bunch of conflicts and new branches have been proposed to complete the RDNN migration

PR #258, opened 2017-12-20
Partial RDNN

mdh34:
Partially change to RDNN without breaking user settings / shortcuts


src/config/GSettingsEngine.vala
```diff
@@ -137,7 +137,7 @@ public class GSettingsConfigurationEngine : ConfigurationEngine, GLib.Object {
             cleaned_id = "default";
         cleaned_id = cleaned_id.replace (".", "-");
 
-        return "org.pantheon.photos.%s.%s".printf (domain, cleaned_id);
+        return "io.elementary.photos.%s.%s".printf (domain, cleaned_id);
```
danrabbit: Is this right? This looks like a gsettings thing which we'd want to not touch :grimacing:

mdh34: changed!

danrabbit:
Pulled and built. One thing I noticed was that this does create an `io.elementary.photos` in `~/.local/share/` instead of using the existing `pantheon-photos` directory. It seems like we just auto-import and it fixes itself, but I'm not sure how other people will feel about this being done in a regular update.

Other than it seems fine to me, but I'd like to get an approve from @codygarver since he's said he's had some trouble with packaging some of the RDNN branches

davidmhewitt:
I'm -1 on changing the data directory. As, @danrabbit mentions above, Photos will auto re-import photos. However, this also has the side effect of potentially losing a bunch of the user's data.

If you don't have the "Write metadata to files" option turned on, any extra metadata you add to photos in your library is stored in the database and not the photo file on disk. So any tags, comments, titles, custom exposure dates, orientation changes etc... are in the database.

I'm pretty sure re-creating the database in a new folder and re-importing all the photos will have the unintended side effect of losing that data.

We'd need to write in a check to see if a database existed in the old location and copy it before trying to create a new one. Or just stick to the old path for now since it's in a hidden folder anyway.

mdh34:
@davidmhewitt in that case let's keep the existing database

PR #259, opened 2017-12-21
Combine close and on_close

danrabbit:
Because like.. why tho

### DIFF
--- a/src/FullScreenWindow.vala
+++ b/src/FullScreenWindow.vala
@@ -56,7 +56,7 @@ public class FullscreenWindow : PageWindow {
 
         var close_button = new Gtk.ToolButton (img, null);
         close_button.tooltip_text = _("Leave fullscreen");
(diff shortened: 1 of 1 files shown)


src/FullScreenWindow.vala
```diff
@@ -135,23 +135,18 @@ public class FullscreenWindow : PageWindow {
...
-                on_close ();
+                close ();
                 return true;
         }
 
         return (base.key_press_event != null) ? base.key_press_event (event) : false;
     }
 
-    private void on_close () {
+    public new void close () {
```
tintou: I would avoid redefining a `close` function but use another name instead

danrabbit: I guess thinking about this there are probably other calls to `close` in subclasses that we don't know about so maybe better to not touch it :grimacing:

PR #283, opened 2018-02-02
Remove pantheon-photos-common package

tintou:


codygarver:
This also does not successfully build because of locale files in /usr/share

codygarver:
Superseded elsewhere.

PR #284, opened 2018-02-15
Always use dark theme

### LINKED ISSUE #31: Use dark theme if available [$5]
Like all multimedia apps are doing, pantheon-photos should use dark theme.

Launchpad Details: [#LP1292340](https://bugs.launchpad.net/bugs/1292340) Erasmo Marín - 2014-03-14 04:06:02 +0000

cassidyjames:
Fixes #31

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -96,7 +96,7 @@ public class Application : Granite.Application {
             command_line.connect (on_command_line);
         }
 
-        Gtk.Settings.get_default().set("gtk-application-prefer-dark-theme", true);
+        Gtk.Settings.get_default ().gtk_application_prefer_dark_theme = true;
         activate.connect (on_activated);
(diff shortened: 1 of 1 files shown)


src/Application.vala
```diff
@@ -97,6 +96,7 @@ public class Application : Granite.Application {
             command_line.connect (on_command_line);
         }
 
+        Gtk.Settings.get_default().set("gtk-application-prefer-dark-theme", true);
```
artemanufrij: perhaps we should use
`Gtk.Settings.get_default ().gtk_application_prefer_dark_theme = true`

cassidyjames: Yep, probably! I was just moving the existing line, but we usually prefer setting the property directly instead of using the setter.

cassidyjames:
This is ready for review.

PR #291, opened 2018-02-22
Page.vala: make view public

danrabbit:
All of this to not duplicate a property :sob:


jeremypw:
You will need to make FullscreenPhotoPage properties construct only as well.

PR #301, opened 2018-03-19
Use Open In, move File Manager

danrabbit:
This changes the menu item label from "Open With..." to "Open In" to match Code and Files and moves the "Show in File Manager" menu item into the "Open In" submenu.

screenshot from 2018-03-19 09 21 54 2x

Removed the sensitivity check since we'll always have "Open In → File Manager"

While we're here, remove some unused string constants


jeremypw:
The expander associated with the "Open in" item is small and dim compared with the following item (although this is also the case in master, if the fix is simple it could be included here).

There is a fair amount of repeated code - can any of this be moved to the common ancestor class (Page.vala)?

jeremypw:
I would question whether "Open in" (or "with") is appropriate with respect to image files/file manager.  The analogous thing with browsers/downloads is "Show in Folder"

cassidyjames:
@danrabbit looks like we have the giant-ass Firefox issue here as well. Might as well set the pixel size in the menu while we're here?

danrabbit:
@jeremypw Fixed!

I agree this should be abstracted, but it looks like that's gonna take some extra work that I'd like to do in another branch

We discussed the naming with Code and decided "Open In" was more flexible since depending on the app this menu might include things like "New Tab" or "New Window" as well as a list of apps

@cassidyjames fixed :)

PR #343, opened 2018-07-25
LibraryWindow: don't mess with focus

### LINKED ISSUE #274: Search box loses focus after character input
#### Steps to Reproduce
1. Focus the search input
2. Type a character

#### Expected Result
The input is still focused.

#### Actual Result
Focus is lost and I have to click back into the search box to continue typing.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/54359381-search-box-loses-focus-after-character-input?utm_campaign=plugin&utm_content=tracker%2F60550306&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60550306&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #274 

If the user wants to change focus, they can do so without our help


src/library/LibraryWindow.vala
```diff
@@ -169,7 +169,6 @@ public class LibraryWindow : AppWindow {
         // setup search bar and add its accelerators to the window
         search_entry = new SearchFilterEntry ();
         search_entry.valign = Gtk.Align.CENTER;
-        search_entry.activate.connect (() => {get_current_page ().grab_focus (); });
```
jeremypw: You should not remove this - it has accessibility implications (cannot transfer keyboard focus to view by pressing Enter) and is not the root cause of the problem.

src/library/LibraryWindow.vala
```diff
@@ -1185,10 +1184,6 @@ public class LibraryWindow : AppWindow {
         }
 
         set_metadata_sidebar_visible (is_metadata_sidebar_visible ());
-
-        if (get_current_page () != null) {
-            get_current_page ().grab_focus ();
-        }
     }
```
jeremypw: This may have accessibility implications - under most circumstances the welcome page should be focussed.
I suggest instead adding another parameter `bool grab_focus = true` and then
```
        if (get_current_page () != null && grab_focus) {
            get_current_page ().grab_focus ();
        }
```

and then  altering the call to this function from `update_view_filter_message ()`  accordinglingy

```
window.toggle_welcome_page (false, "", "", false, false);
```
This will avoid unrelating consequences

danrabbit: I think that is overcomplicating things. We don't try to move focus around like this in any other app. If the user has explicitly chosen to focus some widget, we shouldn't be trying to move around focus under their feet.

jeremypw: By pressing Enter in the search the user is intending to end the search so I think the expected behaviour is to refocus the view.

I'll have to investigate the circumstances of showing the welcome page further but that wasn' the subject of the issue.

src/library/LibraryWindow.vala
danrabbit: You can always move focus with the `tab` key. I don't think we have other apps where we use `enter` as the pattern for moving focus

PR #347, opened 2018-07-27
Spelling: Ellipsis, new camera locked sentence

comradekingu:
There were two extra whitespaces in the sentence, so much improved imo.

### DIFF
--- a/src/camera/ImportPage.vala
+++ b/src/camera/ImportPage.vala
@@ -938,7 +938,7 @@ public class ImportPage : CheckerboardPage {
     // TODO: replace this with approved text for "talking to camera,
     // please wait" once new strings are being accepted.
     protected override string get_view_empty_message () {
-        return _ ("Starting import, please wait...");
+        return _ ("Starting import, please wait…");
     }
 
     private static int64 preview_comparator (void *a, void *b) {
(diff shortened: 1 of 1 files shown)


src/camera/ImportPage.vala
```diff
@@ -1113,7 +1113,7 @@ public class ImportPage : CheckerboardPage {
                     unmount_camera (mount);
                 }
             } else {
-                string locked_message = _ ("The camera is locked by another application.  Photos can only access the camera when it's unlocked.  Please close any other application using the camera and try again.");
+                string locked_message = _ ("Photos can only access the camera when it is not locked by another application. Please close all ither applications using the camera and try again.");
```
BaptisteGelez: Little typo: "ither"

jeremypw:
There is no point in touching ellipsises in the comments - they are not user facing. I would leave them alone.

This should be merged before a string freeze which may be soon.

comradekingu:
@BaptisteGelez If many conventions exist, and the formatting is wrong, it is more likely to end up in something user-facing. I just do it where I see it.

jeremypw:
You are probably right - and it simplifies delinting.  However, looking again at the actual comments, the dots do not seem to be needed at all in this case. Moreover your replaced them with an ellipsis character enclosed by spaces which is a different convention.

comradekingu:
@jeremypw I don't know what it is supposed to be, feel free to change it around. My problem is fixed.

jeremypw:
Not worth worrying about really.  But I have checked and there is a string freeze in operation at the moment so I cannot merge this at the moment anyway.  Will look again after Juno release.

tintou:
@danrabbit @cassidyjames @codygarver there are some string changes in some places so maybe this can rise a new exception?

cassidyjames:
@comradekingu can you check the conflicts here?

PR #350, opened 2018-08-02
AppWindow: use GLib.Action for Quit shortcut

danrabbit:
Gtk.Action is deprecated, start the switch

### DIFF
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -46,7 +46,20 @@ public abstract class AppWindow : PageWindow {
 
     protected GLib.Settings window_settings;
 
+    public const string ACTION_PREFIX = "win.";
+    public const string ACTION_QUIT = "action_quit";
+
+    private const ActionEntry[] action_entries = {
+        { ACTION_QUIT, on_quit }
+    };
+
     construct {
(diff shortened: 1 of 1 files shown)


src/AppWindow.vala
```diff
@@ -44,7 +44,20 @@ public abstract class AppWindow : PageWindow {
...
+        { ACTION_QUIT, on_quit }
+    };
+
     construct {
+        var actions = new SimpleActionGroup ();
+        actions.add_action_entries (action_entries, this);
+        insert_action_group ("win", actions);
+
+        Application.get_instance ().set_accels_for_action (ACTION_PREFIX + ACTION_QUIT, {"<Ctrl>Q"});
+
```
jeremypw: Is there a downside to setting ACTION_PREFIX = "win" and using `ACTION_PREFIX + "." + ACTION_QUIT` to avoid repetition of "win" when inserting the action group?

danrabbit: :man_shrugging: I was just following the convention we've used in other apps. I'm not opposed to either of those alternatives

jeremypw: OK, better be consistent then.

PR #351, opened 2018-08-02
Remove abstracted set_default_title method

danrabbit:
Having the string in a constant is plenty abstracted for maintenance purposes. We don't need to abstract an obvious one-line method

### DIFF
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -60,7 +60,7 @@ public abstract class AppWindow : PageWindow {
         header.set_show_close_button (true);
         this.set_titlebar (header);
 
-        set_default_title ();
+        title = _(Resources.APP_TITLE);
 
         var css_provider = new Gtk.CssProvider ();
         css_provider.load_from_resource ("io/elementary/photos/application.css");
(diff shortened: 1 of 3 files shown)


src/AppWindow.vala
```diff
@@ -58,7 +58,7 @@ public abstract class AppWindow : PageWindow {
         header.set_show_close_button (true);
         this.set_titlebar (header);
 
-        set_default_title ();
+        title = _(Resources.APP_TITLE);
```
jeremypw: Resources.APP_TITLE is already set as translated `APP_TITLE = _("Photos")`?  Or should this be  `APP_TITLE = N_("Photos")`?

danrabbit: To be honest I'm not sure why this is like this, I was just trying to preserve the existing code. Maybe someone more experienced with gettext has an idea why this would be done this way and if it's just wrong

donadigo: I think @jeremypw is right here, the constant should call `N_` and this should be just plain assignment `title = Resources.APP_TITLE;`

jeremypw: I think it still needs to be `title = _(Resources.APP_TITLE);` so that translation subsitution occurs at runtime.  `N_()` marks the string to be translatable.  I think it was @tintou advised me that constant strings should be treated like this.

danrabbit: Ah I think maybe I understand what this discussion is about. Yes I think we should probably make that constant a `N_` and with a translator comment that it's the name of the app

PR #356, opened 2018-08-08
SliderAssembly.vala: Replace deprecated gtk_timeout_foo

danrabbit:
Replace these deprecated properties with their default values as defined here: https://developer.gnome.org/gtk2/stable/GtkSettings.html#GtkSettings--gtk-timeout-initial


danrabbit:
@donadigo They're not being set anywhere, so I can only assume it's using the defaults (which are deprecated)

PR #363, opened 2018-08-09
ImportPage.vala: Replace deprecated render_icon

danrabbit:


jeremypw:
For this rarely used code just using a scale of 1 is probably good enough.

PR #369, opened 2018-08-11
Use GLib.Action for Export

danrabbit:
Theoretically we should be able to use `action_name` for menuitems, but for some reason that doesn't seem to work in practice. So just doing things the way they're already done works :man_shrugging:


jeremypw:
The short-cut does not work - it shows an Emoji chooser for me. Changing it to `<Ctrl><Alt>E` worked for me.

Using the AppWindow action group seems less natural in this case - it doesn't seem to work properly for a ScrolledWindow.  Why not use Glib.Action and an action-group based on an ancestor of the widget? If it is based on e.g. Page.vala then presumably both Collection.Page and LibraryPage could use it - there is duplication at the moment.

danrabbit:
hm I'm unable to reproduce the issue with the emoji picker. I can reproduce an issue however where switching to another view and then back to the main library view makes the shortcut stop working both in this branch and in master. Is that what you're experiencing?

I think you're right. It looks like the current codebase is adding these actions to an action group in Page.vala. It's a little hard to follow along sometimes :sweat_smile:. Will fix

jeremypw:
It turns out that `<Ctrl><Shift>e` is the default ibus shortcut for entering emojis so if you have ibus running and have not changed or disabled this shortcut it will clash with the export action shortcut in Photos.

PR #372, opened 2018-08-15
TextEntryDialog: Rewrite without UI files

danrabbit:
Rewrites the dialog with GObject-style construction (almost) and no UI files or Gtk.Builder or anything.

Solves a few deprecation warnings

### DIFF
--- a/src/Dialogs/TextEntryDialog.vala
+++ b/src/Dialogs/TextEntryDialog.vala
@@ -29,8 +29,14 @@ public class TextEntryDialog : Gtk.Dialog {
     private unowned OnModifyValidateType on_modify_validate;
     private Gtk.Entry entry;
 
-    public TextEntryDialog (OnModifyValidateType? modify_validate, string title, string label,
-string? initial_text, Gee.Collection<string>? completion_list, string? completion_delimiter) {
+    public TextEntryDialog (
+        OnModifyValidateType? modify_validate,
+        string title,
+        string label,
+        string? initial_text,
+        Gee.Collection<string>? completion_list,
+        string? completion_delimiter
+    ) {
         Object (
             completion_delimiter: completion_delimiter,
             completion_list: completion_list,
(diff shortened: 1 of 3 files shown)


src/Dialogs/TextEntryDialog.vala
```diff
@@ -19,51 +19,62 @@
...
-    private Gtk.Button button1;
-    private Gtk.Button button2;
-    private Gtk.ButtonBox action_area_box;
 
-    public void set_builder (Gtk.Builder builder) {
-        this.builder = builder;
-    }
+    public TextEntryDialog (OnModifyValidateType? modify_validate, string title, string label,
+string? initial_text, Gee.Collection<string>? completion_list, string? completion_delimiter) {
+        Object (
```
jeremypw: Align parameters under opening bracket?

src/Dialogs/TextEntryDialog.vala
```diff
@@ -19,51 +19,62 @@
...
+    construct {
+        var name_label = new Gtk.Label (label);
+        name_label.halign = Gtk.Align.START;
+        name_label.get_style_context ().add_class (Granite.STYLE_CLASS_H4_LABEL);
 
-        entry = builder.get_object ("entry") as Gtk.Entry;
-        entry.set_text (initial_text != null ? initial_text : "");
+        entry = new Gtk.Entry ();
+        entry.hexpand = true;
+        entry.text = initial_text != null ? initial_text : "";
```
jeremypw: `entry.text = initial_text ?? ""` is more concise.

jeremypw:
Note: the alignment example did not work in the review comment - not WYSIWYG :(

danrabbit:
Yeah I think we should probably document an official style for how to deal with this.

My reasoning for using this style is largely based around when we started using linters in other projects and having a single indent for things between brackets on multiple lines is what the linter expects.

I also feel like it's more maintainable if the parameters change. You don't necessarily have to touch a bunch of lines to fix the alignment.

PR #378, opened 2018-08-16
AdjustDateTimeDialog: Use Grids instead of Boxes and Alignments

danrabbit:
Fix a deprecation warning by using Gtk.Grid instead of nested boxes and Gtk.Alignment

This dialog is still a little wonky looking, but I'm just here to fix the warning

## BEFORE
screenshot from 2018-08-16 11 40 45 2x

## AFTER
screenshot from 2018-08-16 11 36 55 2x

### DIFF
--- a/src/Dialogs/AdjustDateTimeDialog.vala
+++ b/src/Dialogs/AdjustDateTimeDialog.vala
@@ -1,6 +1,6 @@
 /*
 * Copyright (c) 2009-2013 Yorba Foundation
-*               2017 elementary  LLC. (https://github.com/elementary/photos)
+*               2017-2018 elementary, Inc. (https://elementary.io)
 *
 * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


src/Dialogs/AdjustDateTimeDialog.vala
```diff
@@ -47,26 +47,29 @@ public class AdjustDateTimeDialog : Gtk.Dialog {
...
+            (_("_Apply")), Gtk.ResponseType.OK
+        );
+
+        deletable = false;
+        modal = true;
+        resizable = false;
+        title = _(Resources.ADJUST_DATE_TIME_LABEL);
+        transient_for = AppWindow.get_instance ();
+        set_default_response (Gtk.ResponseType.OK);
     }
```
jeremypw: Could set properties in an Object () statement in the constructor. Not sure which is preferred.

PR #382, opened 2018-08-21
Save window dimensions on configure event

danrabbit:
Save all settings on window state change and only save window size if the window is not maximized so that we can restore that last non-maximized size


jeremypw:
I am not sure of the benefits (if any) of spamming the settings on every configure event, including those that do not actually change the size of the window? Why was this changed from save on quit?

danrabbit:
We've had trouble previously with only saving changes on quit. If the app is ever forcefully quit changes won't be saved. It's more simple and more reliable to save when things change.

jeremypw:
I would consider that if a program needs to be forcefully quit something has gone wrong and it is questionable whether its state should be saved.  I can see the argument in the case of something important like changes to a document but not in the case of something trivial like window size.

PR #388, opened 2018-08-25
ProgressDialog: rewrite as a Dialog without alignments and boxes

danrabbit:
Solves a deprecation warning

## BEFORE
screenshot from 2018-08-24 20 44 39 2x

## AFTER
screenshot from 2018-08-24 20 43 48 2x


danrabbit:
Yes moving it was intentional :)

PR #392, opened 2018-08-26
AdjustTool: Color tint and temperature scales

danrabbit:
screenshot from 2018-08-26 10 41 14 2x


danrabbit:
Even if we did hardcode the palette, this isn't going to look good in Adwaita:
screenshot from 2018-08-27 08 30 08 2x

PR #397, opened 2018-08-28
LibraryPhotoPage: Clean up action entries

danrabbit:
Same as #395 

As far as I can tell all of these shortcuts are broken in master, so I didn't introduce that in this branch


danrabbit:
Yes sorry, forgot to mention that. There are a couple of actions that were meant to open menus that don't exist anymore

PR #404, opened 2018-08-30
AppWindow.vala: Move more things into construct

danrabbit:
* Get rid of the whole `build_headerbar` thing
* Add a bunch of stuff to `construct` and put it in order
* Cache application instance

LibraryWindow:
* Move the `build_headerbar` stuff to `construct` since that method doesn't exist to override anymore
* Use property binding for title


src/AppWindow.vala
```diff
@@ -132,19 +103,43 @@ public abstract class AppWindow : PageWindow {
...
         header.pack_end (redo_btn);
         header.pack_end (undo_btn);
+
+        icon_name = "multimedia-photo-manager";
+        title = _(Resources.APP_TITLE);
+
+        add_action_entries (action_entries, this);
+        set_titlebar (header);
+
+        Application.get_instance ().set_accels_for_action (ACTION_PREFIX + ACTION_FULLSCREEN, {"F11"});
```
tintou: It'd be nice to also store the result of `Application.get_instance ()` as it is reused multiple times

src/library/LibraryWindow.vala
```diff
@@ -143,6 +143,35 @@ public class LibraryWindow : AppWindow {
...
+        var settings = new Gtk.MenuButton ();
+        settings.image = new Gtk.Image.from_icon_name ("open-menu", Gtk.IconSize.LARGE_TOOLBAR);
+        settings.tooltip_text = _("Settings");
+        settings.popup = settings_menu;
+        settings.show_all ();
+
+        header.pack_end (settings);
+        header.set_custom_title (top_display);
+
+        notify["title"].connect (() => {
```
tintou: probably use `bind_property` here?

PR #429, opened 2018-09-11
FullScreenWindow: GObject-style

danrabbit:


jeremypw:
This feels odd.  The parent class has a public setter function for "current page" but here you have added a construct only property whose only purpose seems to be to allow GObject style construction. Is it worthwhile?

I investigated making `current_page` a property of the parent class but it had a lot of repercussions (and I thought Files code was a mess!)

jeremypw:
Another possible problem is that after construction the property 'page' may no longer be valid because the current page of PageWindow can change (at least theoretically). 

Perhaps you should make the getter private and rename it something more specific like "initial_page".

PR #440, opened 2018-09-18
Move all Pages to Views folder

danrabbit:
Hopefully making it easier to find subclasses of Page and its subclasses


tintou:
Somebody is not updating the POTFILES :thinking:

PR #443, opened 2018-09-18
Split up SearchDialog classes into separate files

danrabbit:
these classes are too complex to all be embedded in a single file even though they don't technically need to be public


tintou:
@danrabbit Please rebase :)

PR #448, opened 2018-09-27
Make AbstractSearchRow a Gtk.Grid subclass

danrabbit:
* No need for a `get_widget` method when this is a widget
* Set grid spacing  and the override for `parent` in the abstract class instead of in each class


jeremypw:
Looks sane, but how do you access the saved search dialog?  I cannot seem to find it in the interface.

danrabbit:
@jeremypw It's super hidden. Perform a search and then hit Ctrl + S

jeremypw:
OK, thanks.  I suppose it makes some sense, but I would expect an entry in the context menu of the view or the search entry.    Also the dialog is not prefilled with the actual search you want to save. However, those are separate issues.

Issue #460, opened 2018-10-23
Google Photos login in a tiny box (screenshot)


linkert:
Wanted to publish photos to my Google Photos account through the build in function in Photos but I could not get past the login screen as It was a miniature box.
![screenshot from 2018-10-23 15-45-21](https://user-images.githubusercontent.com/44402757/47365108-21ea8680-d6db-11e8-8d2c-7c4cf3c2e06e.png)

PR #461, opened 2018-10-23
fix scrolledwindow not expanding

arshubham:
Fixes https://github.com/elementary/photos/issues/460 .
Fixed facebook's and yandex's too.


tintou:
Thanks for your contribution, hopefully this UI will arrive in the Online Accounts pane of the settings application :)

arshubham:
Looking forward to it.

PR #482, opened 2019-01-11
Add meson post install script

danrabbit:
supersedes #433

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -131,3 +131,5 @@ subdir('po')
 if get_option('documentation')
     subdir('docs')
 endif
+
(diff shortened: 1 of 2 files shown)


worldofpeace:
Was that superseded because I was using `argv`? Honestly felt like that made more sense in general.

danrabbit:
@worldofpeace more because it seems I couldn't update the branch, but this is also the same script as we're using in all the other repos so I figure it'd be a no brainer for someone else to approve it

jtojnar:
The original PR was better in that it worked even when datadir was changed.

Issue #488, opened 2019-02-15
Use FileChooserNative


danrabbit:
- [x] https://github.com/elementary/photos/blob/3dfa73156dc7d2e5f50a133f8d91a385570639e5/src/Dialogs/Dialogs.vala#L92
- [x] https://github.com/elementary/photos/blob/3dfa73156dc7d2e5f50a133f8d91a385570639e5/src/Dialogs/Dialogs.vala#L121
- [x] https://github.com/elementary/photos/blob/43bff10b9228814d07b54637e5b6961f02893221/src/Views/DirectPhotoPage.vala#L415
- [x] https://github.com/elementary/photos/blob/cc0512b046216521d7279222c74c6610ceb4aa9f/src/library/LibraryWindow.vala#L615

peteruithoven:
These all seem switched to native. 

@danrabbit why not include the following one?
https://github.com/elementary/photos/blob/master/src/Dialogs/Dialogs.vala#L607

PR #489, opened 2019-02-16
Use FileChooserNative

danrabbit:
Partially addresses #488

### DIFF
--- a/src/Dialogs/Dialogs.vala
+++ b/src/Dialogs/Dialogs.vala
@@ -86,26 +86,24 @@ public File? choose_file (string current_file_basename) {
     if (current_export_dir == null)
         current_export_dir = File.new_for_path (Environment.get_home_dir ());
 
-    string file_chooser_title = VideoReader.is_supported_video_filename (current_file_basename) ?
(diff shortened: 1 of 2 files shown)


src/Dialogs/Dialogs.vala
```diff
@@ -116,11 +114,15 @@ public File? choose_dir (string? user_title = null) {
...
 
     if (user_title == null)
-        user_title = _ ("Export Photos");
-
-    Gtk.FileChooserDialog chooser = new Gtk.FileChooserDialog (user_title,
-            AppWindow.get_instance (), Gtk.FileChooserAction.SELECT_FOLDER, _ ("_Cancel"),
-            Gtk.ResponseType.CANCEL, _ ("_Select"), Gtk.ResponseType.ACCEPT, null);
+        user_title = _("Export Photos");
+
+    var chooser = new Gtk.FileChooserDialog (
```
jeremypw: Should be Gtk.FileChooserNative  (else Cancel button missing)

danrabbit: Oops, silly mistake

PR #490, opened 2019-02-16
LibraryWindow: Use FileChooserNative

danrabbit:


src/library/LibraryWindow.vala
```diff
@@ -612,19 +612,21 @@ public class LibraryWindow : AppWindow {
...
         import_dialog.set_current_folder (import_dir);
 
-        int response = import_dialog.run ();
-
-        if (response == Gtk.ResponseType.OK) {
+        if (import_dialog.run () == Gtk.ResponseType.ACCEPT) {
             /* Set invisible the dialog because some wm keep it in front
              * giving the sensation of a froozen program.*/
-            import_dialog.set_visible (false);
+            import_dialog.visible = false;
```
peteruithoven: This seems like a weird exception I don't see in a other similar PR's, like for Videos, Code and Music.

I did notice this comment: https://github.com/elementary/code/pull/643/files#diff-4089e5f35cbce620c0c1758fb579b361R764
Could that be a better way? Although it seems a bit weird that the `get_uris ()` is then called after the `destroy`? Maybe it would be better to save the uri in a var before calling `destroy`?

PR #502, opened 2019-04-09
Dialogs: revert_editable_dialog as Granite.MessageDialog

danrabbit:


jeremypw:
Not sure how to trigger this.  When I choose "Revert to original" from the context menu there is no dialog.

ryonakano:
## BEFORE

Screenshot from 2019-08-13 12-07-46

## AFTER

Screenshot from 2019-08-13 12-09-01

PR #498, opened 2019-03-16
Fix building against latest gexiv2

worldofpeace:
Fixes https://github.com/elementary/photos/issues/484

PR #505, opened 2019-04-16
photos/PhotoMetadata: fix missing line accidentally dropped with PR #498

decathorpe:
PR #498 accidentally dropped an unrelated line completely when wrapping the GExiv2 code inside a conditional. This PR re-introduces the missing line.


decathorpe:
Thanks!

PR #513, opened 2019-07-26
Drop styles that are already in system stylesheet

cassidyjames:
It looks like we were carrying some custom styles from when Photos was a light app, but we wanted the canvas behind images to be dark. Now that we're using the dark style, we can drop a lot. Plus, checkerboard and temperature scales are in the system stylesheet.

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -22,6 +22,24 @@
     color: @text_color;
 }
 
+scale.temperature trough {
+    background-image:
+        linear-gradient(
+            to right,
+            @BLUEBERRY_500,
+            @BANANA_500
+        );
+}
+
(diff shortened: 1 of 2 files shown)


data/application.css
```diff
@@ -76,24 +22,6 @@
     color: @text_color;
 }
 
-scale.temperature trough {
```
danrabbit: Can we leave these in? This is actually a different style than what's in the stylesheet right now

cassidyjames: Sure, I thought the difference was because we did Photos first and figured we'd want to rely on the system stylesheet here. But I'll add it back.

PR #520, opened 2019-07-31
Improve export dialog UI

### LINKED ISSUE #517: Pixels field title in export window not aligned to the left
Right click on photo > Other Actions > Export...
Screenshot from 2019-07-30 23 09 28
It's not capitalized and doesn't have a colon. The pixels field itself might have to be thought out, as it's the same size as the selection fields. I know this is another UI element taken from Shotwell, so thought I'd bring it up if anyone felt it needed to be redesigned. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/77993400-pixels-field-title-in-export-window-not-aligned-to-the-left?utm_campaign=plugin&utm_content=tracker%2F60550306&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60550306&utm_medium=issues&utm_source=github).
</bountysource-plugin>

fabiozaramella:
Fixes #517 

Old:
Old
New:
Screenshot from 2019-08-04 12 30 59@2x

### DIFF
--- a/src/Dialogs/ExportDialog.vala
+++ b/src/Dialogs/ExportDialog.vala
@@ -100,7 +100,7 @@ public class ExportDialog : Gtk.Dialog {
 
         Gtk.Label size_label = new Gtk.Label.with_mnemonic (_("_Size in pixels:"));
         size_label.halign = Gtk.Align.END;
-        quality_label.use_underline = true;
+        size_label.use_underline = true;
         size_label.mnemonic_widget = pixels_entry;
 
         export_metadata = new Gtk.CheckButton.with_label (_("Export metadata"));
@@ -142,10 +142,11 @@ public class ExportDialog : Gtk.Dialog {
         format_combo.changed.connect (on_format_changed);
         pixels_entry.changed.connect (on_pixels_changed);
         pixels_entry.activate.connect (() => {
-            if ((pixels_entry.get_text_length () > 0) && (int.parse (pixels_entry.get_text ()) > 0))
+            if ((pixels_entry.get_text_length () > 0) && (int.parse (pixels_entry.get_text ()) > 0)) {
                 response (Gtk.ResponseType.OK);
-            else
+            } else {
                 pixels_entry.set_value (current_scale);
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


cassidyjames:
Definitely an improvement. I wonder: while we're here, would it make sense to switch the quality combo to a scale, and the size entry to a spinbutton?

fabiozaramella:
@cassidyjames I already thought about switching to a spinbutton, but i didn't changed it because it requires a range of values and i wasn't sure about the proper number to put it (the current entry accept a maximum of 6 digits anyway so...). Should i update this PR or make a different one?

src/Dialogs/ExportDialog.vala
```diff
@@ -94,13 +93,15 @@ public class ExportDialog : Gtk.Dialog {
...
+        pixels_entry = new Gtk.SpinButton.with_range (1, 999999, 1);
+        pixels_entry.set_max_length (6);
+        pixels_entry.set_value (current_scale);
+        pixels_entry.set_digits (0);
 
-        Gtk.Label pixels_label = new Gtk.Label.with_mnemonic (_("_pixels"));
-        pixels_label.mnemonic_widget = pixels_entry;
+        Gtk.Label size_label = new Gtk.Label.with_mnemonic (_("_Size in pixels:"));
+        size_label.halign = Gtk.Align.END;
+        quality_label.use_underline = true;
```
danrabbit: Is this a copy/paste mistake or is this property just in the wrong spot :)

src/Dialogs/ExportDialog.vala
```diff
@@ -139,9 +141,11 @@ public class ExportDialog : Gtk.Dialog {
         constraint_combo.changed.connect (on_constraint_changed);
         format_combo.changed.connect (on_format_changed);
         pixels_entry.changed.connect (on_pixels_changed);
-        pixels_entry.insert_text.connect (on_pixels_insert_text);
         pixels_entry.activate.connect (() => {
-            response (Gtk.ResponseType.OK);
+            if ((pixels_entry.get_text_length () > 0) && (int.parse (pixels_entry.get_text ()) > 0))
```
danrabbit: Can we use brackets for if/else please

src/Dialogs/ExportDialog.vala
fabiozaramella: Oops, copy/past mistake

src/Dialogs/ExportDialog.vala
fabiozaramella: done :+1:

Issue #97, opened 2017-04-18
Missing SVG viewing/scaling support


elementaryBot:
Even though vector graphics aren't photos and some editing features aren't applicable SVG support (à la Eye of GNOME, gThumb, Gwenview etc.) would come in handy given that "Photos" is the default image viewer of elementary OS.

Launchpad Details: [#LP1423312](https://bugs.launchpad.net/bugs/1423312) WKraven - 2015-02-18 19:29:30 +0000

danrabbit:
![screenshot from 2018-03-12 22 42 42 2x](https://user-images.githubusercontent.com/7277719/37324462-c1244b1e-2646-11e8-9c2a-2599f5392f50.png)

I can confirm that Photos doesn't currently open SVG files

codygarver:
There's a bounty for this issue [on Bountysource](https://www.bountysource.com/issues/56143083-missing-svg-viewing-scaling-support).

rucuriousyet:
I wonder if maybe we couldn't just convert the SVG to a PNG on the fly for the size of the window.... It might be a bad idea for complex vectors, but that might be a decent quick solution since converting a SVG to PNG with something like Cairo and RSvg is very quick

danirabbit:
I dunno I think I'm going to close this as "out of scope". This is similar imo to the request to support .psd files. There are dedicated apps for these kinds of file formats. Photos is really for managing your camera roll

PR #522, opened 2019-08-13
Add .SVG support

oct2pus:
Should address issue #97 
I'm not finished but I at least have a proof of concept done. I mostly wanted to get feedback and find out if this is still a feature that Elementary is interested in. 

image

Problems:
I need to remove SVGs from the list of exportable files (it exports to PNG), and currently resizing resizes the converted image, instead of using the actual SVG file to provide a higher quality zoom. I have done zero work for metadata.


cassidyjames:
I definitely think it's useful to be able to at least preview SVGs. Even if it's not a full vector support with zoom, I think it's better than nothing.

How do edits work? Does it apply them to the converted image as well? Would it be reasonable to just make the controls insensitive for SVG (and find a way to let the user know why) instead of trying to deal with editing the converted image?

oct2pus:
> I definitely think it's useful to be able to at least preview SVGs. Even if it's not a full vector support with zoom, I think it's better than nothing.

Good to hear!

> How do edits work? Does it apply them to the converted image as well? Would it be reasonable to just make the controls insensitive for SVG (and find a way to let the user know why) instead of trying to deal with editing the converted image?

The SVG files are (currently) read as PNGs through Pixbuf.read and can be modified as like other PNG files, Unless I missed something glaring then this does not actually affect the underlying SVG file. I suspect getting better resizing support will break modifications on a resize (or entirely if I can find a way to render to SVG directly without a major code rewrite). I'd personally prefer to get more dynamic sizing working and then disable modifications if it comes to that, but considering the purpose of the application just treating them like PNG files might be preferable?

cassidyjames:
@elementary/ux thoughts on disabling modifications on vector formats?

ryonakano:
Do we want to set Photos as a default app to open .svg files while we're here?

oct2pus:
For the record I'm still waiting for a design team consideration before moving forward on this

adding properly scaling SVGs is going to be harder than I thought tho, I was struggling about a month ago to figure out where resizing actually occurs in the code?

danrabbit:
Yeah I think editing SVGs is probably outside the scope of Photos. I think it would make sense for them to be read only :+1:

jeremypw:
@oct2pus Looks like you got the go-ahead from the design team for this - do you have time to work on it?

jeremypw:
@danrabbit  I am not sure there is a convincing use case for this if the svg images are read-only.  Isn't Photos intended for managing albums of digital photos?  The code is bloated enough as it is.

Converting to draft as there are conflicts and no recent activity.

jeremypw:
Closing as no response from author.  Please re-open if you wish to continue working on this.

PR #525, opened 2019-08-13
Open in menu in photo viewer

### LINKED ISSUE #340: Add "Open in..." menu to Photo Viewer
Photos has "Open in..." context menu and we can open applications which can handle images.

But the context menu of photo viewer is different from Photo's one and has only "File Manager" menu.

I would say that it is even better to add the same "Open in..." menu which is used in Photos to photo viewer.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61259033-add-open-in-menu-to-photo-viewer?utm_campaign=plugin&utm_content=tracker%2F60550306&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60550306&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ryonakano:
VirtualBox_elementary OS_13_08_2019_19_07_44

Fixes #340

~[The code from line 621 to 708 in this PR](https://github.com/ryonakano/photos/blob/4d3ce294b2990adfe9a9a81ff6027a73a181662f/src/Views/DirectPhotoPage.vala#L192-L283) is essentially the same with [from line 192 to line 283 of `src/Views/LibraryPhotoPage.vala`](https://github.com/elementary/photos/blob/d21151a284549c6a019259456f577c93810dbe29/src/Views/LibraryPhotoPage.vala#L621-L708) (just fixed to follow the elementary style coding), so this PR might not be effective… :cry:~
Edit: I guess we can immigrate the duplicated functions into the `EditingHostPage.vala` which those classes inherit in another PR.


Fatih20:
Looking for this as well. It's annoying when I need togo to file manager to edit the photo in GIMP or something.

src/Views/DirectPhotoPage.vala
```diff
@@ -171,6 +192,97 @@ public class DirectPhotoPage : EditingHostPage {
...
+            AppWindow.get_instance ().set_busy_cursor ();
+            get_photo ().open_with_external_editor (app);
+            AppWindow.get_instance ().set_normal_cursor ();
+        } catch (Error err) {
+            AppWindow.get_instance ().set_normal_cursor ();
+            open_external_editor_error_dialog (err, get_photo ());
+        }
+    }
+
+    private void on_open_with_raw (string app) {
```
tintou: Can you change it so that you get the command_line at the very end, in the future, it'd be even better if we could use that object to spawn the command with arguments
```suggestion
    private void on_open_with_raw (AppInfo app) {
```

src/Views/DirectPhotoPage.vala
```diff
@@ -171,6 +192,97 @@ public class DirectPhotoPage : EditingHostPage {
         return true;
     }
 
+    private void populate_external_app_menu (Gtk.Menu menu, PhotoFileFormat file_format, bool raw) {
+        SortedList<AppInfo> external_apps;
+        string[] mime_types;
+
+        foreach (Gtk.Widget item in menu.get_children ()) {
```
tintou:
```suggestion
        foreach (weak Gtk.Widget item in menu.get_children ()) {
```

src/Views/DirectPhotoPage.vala
```diff
@@ -171,6 +192,97 @@ public class DirectPhotoPage : EditingHostPage {
...
+                } else {
+                    on_open_with (app.get_commandline ());
+                }
+            });
+            menu.add (item_app);
+        }
+        menu.show_all ();
+    }
+
+    private void on_open_with (string app) {
```
tintou:
```suggestion
    private void on_open_with (AppInfo app) {
```
Use it here too :)

src/Views/DirectPhotoPage.vala
```diff
@@ -171,6 +192,97 @@ public class DirectPhotoPage : EditingHostPage {
...
+
+    private void on_open_with_raw (AppInfo app) {
+        if (!has_photo ()) {
+            return;
+        }
+
+        if (get_photo ().get_master_file_format () != PhotoFileFormat.RAW) {
+            return;
+        }
+
```
tintou:
```
var app_window = AppWindow.get_instance ();
```
then use it everywhere

PR #531, opened 2019-09-04
Application: Use Photos namespace

danrabbit:
And replace Photos.Application.get_instance () with GLib.Application.get_default ()

### DIFF
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -88,7 +88,7 @@ public abstract class AppWindow : PageWindow {
         add_action_entries (action_entries, this);
         set_titlebar (header);
 
-        var application_instance = Application.get_instance ();
+        var application_instance = ((Photos.Application) GLib.Application.get_default ());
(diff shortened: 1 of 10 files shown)


src/AppWindow.vala
```diff
@@ -88,7 +88,7 @@ public abstract class AppWindow : PageWindow {
         add_action_entries (action_entries, this);
         set_titlebar (header);
 
-        var application_instance = Application.get_instance ();
+        var application_instance = ((Photos.Application) GLib.Application.get_default ());
```
tintou:
```suggestion
        weak Photos.Application application_instance = ((Photos.Application) GLib.Application.get_default ());
```

src/Application.vala
```diff
@@ -120,7 +120,7 @@ public class Application : Gtk.Application {
      * will trigger this and exit, and we'll need to bring the window back up...
      */
     public static void on_activated () {
-        get_instance ();
+        GLib.Application.get_default ();
```
tintou: this can go away

Issue #501, opened 2019-04-06
Sorting issues


zelikos:
Sorting photos by title works as expected, however sorting by exposure date has problems.

1) The "exposure date" it reads seems to be the date the picture was last modified, rather than when it was actually created.

2) It doesn't properly sort by date, anyway:
![Screenshot from 2019-04-05 20-39-02_modified](https://user-images.githubusercontent.com/26857149/55662617-261ae400-57e3-11e9-8309-400147de31ea.png)

In the above screenshot, my pictures are sorted by exposure date, in *ascending* order, but more *recent* pictures are near the top, and only some of them are in the correct order.

justledbetter:
I'm working on a sort-of fix for this - by adding an option to sort by "File Modification Date".  Getting ready to submit a pull request to see if the Dev team likes it.

PR #563, opened 2020-04-10
Workaround for Issue #501

justledbetter:
Allow the user to sort by File Modification Time (on-disk timestamp) in addition to Exposure Date and Title.  This helps provide a way to get a more predictable linear sort of the library photos.

Still looking into the Exposure Date sort/data issues and will submit a separate PR for that, if I am able to solve it.

I have not updated the translation files in this pull, since I don't know how to do that and don't want to break anything.  Please advise if you would like for me to do that.


justledbetter:
Inviting @cassidyjames to review this change. Hope I did good!  :)

jeremypw (approved):
Thanks for the contribution and sorry that it was overlooked for so long!

This works as expected.  The code style matches the existing codebase rather than current elementary standards but that can be overlooked as the existing code needs an overhaul anyway.

Issue #566, opened 2020-05-11
Library window position state isn't saved/restored


cassidyjames:
It looks like we save the size of the library window, but not the position.

4jNsY6fCVqZv:
Hello, I'd like to work on a solution. Probably the appropriate place is https://github.com/elementary/photos/blob/master/data/io.elementary.photos.gschema.xml, isn't it?
Is there some kind of a code reference in other elementary apps that helps me to make the necessary changes?

cassidyjames:
Yeah, here is the GSchema where we are saving the size of the library window:

https://github.com/elementary/photos/blob/c6b0381baf79e62d1d7cd4bbfd4502da0ecc365f/data/io.elementary.photos.gschema.xml#L155-L165

Here is a recent pull request where someone updated how Code saves its window state: https://github.com/elementary/code/pull/794

It looks like you'd want to update LibraryWindow.vala here:

https://github.com/elementary/photos/blob/b62cc8f0119490ce1f959abc3cbe43be7748e24c/src/library/LibraryWindow.vala#L146-L149

and here:

https://github.com/elementary/photos/blob/b62cc8f0119490ce1f959abc3cbe43be7748e24c/src/library/LibraryWindow.vala#L514-L520

…to save and restore the window position in addition to the width/height.

4jNsY6fCVqZv:
@cassidyjames Should the key name ideally stay with "library-xxx"? I have seen that other apps tend to use something like "window-xxx" instead.

cassidyjames:
@4jNsY6fCVqZv for Photos it looks like this state is specifically for the Library window, i.e. not the previewer window. So the current name makes sense to me. Default sizes are up to the different apps and their own use cases and designs—I don't think it's necessary to have them all be the same default size.

4jNsY6fCVqZv:
@cassidyjames 
> Default sizes are up to the different apps and their own use cases and designs

The differences are so marginal, does that really have something to do with design and use cases? This is not self-explanatory for me yet.

4jNsY6fCVqZv:
@cassidyjames Question for better understanding, after a look into different elementary default apps has confused me a bit, because the solutions sometimes look very different: The current LibraryWindow.vala solution already saves/restores the status of the window size and it's just a matter of adding the position value, right? Or is it also a matter of introducing saving/restoring as such?

cassidyjames:
[...]

PR #569, opened 2020-05-18
Save/restore the status of the library window position

4jNsY6fCVqZv:
Fixes https://github.com/elementary/photos/issues/566

gschema.xml
- [x] Use common key for library window width and height ("library-size").
- [x] Add the initial key for the library window position ("library-position").

LibraryWindow.vala
- [x] Use "library-size".
- [x] Initially add the "library-position".

appdata.xml
- [ ] Update the version of Photos to 2.7.1

### DIFF
--- a/src/library/LibraryWindow.vala
+++ b/src/library/LibraryWindow.vala
@@ -144,8 +144,8 @@ public class LibraryWindow : AppWindow {
 
     construct {
         set_default_size (
-            window_settings.get_int ("library-position"),
-            window_settings.get_int ("library-size")
+            window_settings.get ("library-position", "(ii)", out window_x, out window_y);
+            window_settings.get ("library-size", "(ii)", out width, out height);
         );
 
         if (window_settings.get_boolean ("library-maximize")) {
@@ -513,8 +513,8 @@ public class LibraryWindow : AppWindow {
 
     protected override void on_quit () {
         window_settings.set_boolean ("library-maximize", is_maximized);
-        window_settings.set_int ("library-position", dimensions.position);
-        window_settings.set_int ("library-size", dimensions.size);
+        window_settings.set ("library-position", "(ii)", out window_x, out window_y);
+        window_settings.set ("library-size", "(ii)", out width, out height);
 
         base.on_quit ();
     }
--- a/src/library/LibraryWindow.vala
+++ b/src/library/LibraryWindow.vala
@@ -144,8 +144,8 @@ public class LibraryWindow : AppWindow {
 
     construct {
         set_default_size (
-            window_settings.get_int ("library-width"),
-            window_settings.get_int ("library-height")
+            window_settings.get_int ("library-position"),
+            window_settings.get_int ("library-size")
         );
 
         if (window_settings.get_boolean ("library-maximize")) {
@@ -513,8 +513,8 @@ public class LibraryWindow : AppWindow {
 
     protected override void on_quit () {
         window_settings.set_boolean ("library-maximize", is_maximized);
-        window_settings.set_int ("library-width", dimensions.width);
-        window_settings.set_int ("library-height", dimensions.height);
+        window_settings.set_int ("library-position", dimensions.position);
+        window_settings.set_int ("library-size", dimensions.size);
 
         base.on_quit ();
     }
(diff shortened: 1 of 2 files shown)


src/library/LibraryWindow.vala
```diff
@@ -144,8 +144,8 @@ public class LibraryWindow : AppWindow {
 
     construct {
         set_default_size (
-            window_settings.get_int ("library-width"),
-            window_settings.get_int ("library-height")
+            window_settings.get_int ("library-position"),
+            window_settings.get_int ("library-size")
```
jeremypw, line 148: These settings types are no longer "int".  You need to use the @get () function like this:

`window_settings.get ("window-size", "(ii)", out width, out height)`

src/library/LibraryWindow.vala
```diff
@@ -513,8 +513,8 @@ public class LibraryWindow : AppWindow {
 
     protected override void on_quit () {
         window_settings.set_boolean ("library-maximize", is_maximized);
-        window_settings.set_int ("library-width", dimensions.width);
-        window_settings.set_int ("library-height", dimensions.height);
+        window_settings.set_int ("library-position", dimensions.position);
+        window_settings.set_int ("library-size", dimensions.size);
```
jeremypw, line 517: As above, you need to use the @set () function like this:

`window_settings.set_int ("library-size", "(ii)", width, height);`

4jNsY6fCVqZv:
@jeremypw Thanks for your feedback! Is there anything else to do in this PR besides updating the appdata.xml?

jeremypw:
@4jNsY6fCVqZv You need to check that this actually works - I don't think it will at the moment as you did not copy my code exactly. Always test your code works before submitting for review.

src/library/LibraryWindow.vala
```diff
@@ -513,8 +513,8 @@ public class LibraryWindow : AppWindow {
 
     protected override void on_quit () {
         window_settings.set_boolean ("library-maximize", is_maximized);
-        window_settings.set_int ("library-width", dimensions.width);
-        window_settings.set_int ("library-height", dimensions.height);
+        window_settings.set ("library-position", "(ii)", out window_x, out window_y);
+        window_settings.set ("library-size", "(ii)", out width, out height);
```
jeremypw, line 517: You should not have the "out" keyword when setting - you are not retrieving values out of the settings - you are putting them in.

src/library/LibraryWindow.vala
```diff
@@ -144,8 +144,8 @@ public class LibraryWindow : AppWindow {
 
     construct {
         set_default_size (
-            window_settings.get_int ("library-width"),
-            window_settings.get_int ("library-height")
+            window_settings.get ("library-position", "(ii)", out window_x, out window_y);
+            window_settings.get ("library-size", "(ii)", out width, out height);
```
jeremypw, line 148: You need to rewrite this because  `settings.get ()` does not return a value so cannot be used as parameter for `set_default_size ()`.

This should work  (I havent checked it)
```
int width, height;
window_settings.get ("library-size", "(ii)", out width, out height);
set_default_size (width, height);
```

src/library/LibraryWindow.vala
4jNsY6fCVqZv: window_settings.set_int?

jeremypw:
Refactoring the code does not usually warrant any entry in the release appdata, if it is unnoticeable to the user.

4jNsY6fCVqZv:
@jeremypw can you please delete this draft?

4jNsY6fCVqZv:
> Refactoring the code does not usually warrant any entry in the release appdata, if it is unnoticeable to the user.

i thought that users will notice this when the position of their window is saved and restored.

jeremypw:
Ah, sorry - I thought this was just a refactoring.  You are right in that case - a line in appdata would be warranted.

4jNsY6fCVqZv:
@jeremypw Can you please delete this PR completely as an exception? I'm still learning. And I don't want to mess up the repository.

jeremypw:
As far as I am aware I cannot delete a PR, only close it. As you originated the PR, if anyone can delete it you can.  As long as the corresponding code branch has been deleted so it doesn't really affect the elementary repository.

4jNsY6fCVqZv:
@cassidyjames Maybe a repository admin? I have no permissions in this repository and don't want to clutter it up. I've been playing with the web interface, but I can't handle it. It rather confuses me.

cassidyjames:
@4jNsY6fCVqZv it's closed. :smile: That's all we need to do.

PR #571, opened 2020-06-10
Use granite instead of libunity

### LINKED ISSUE #532: Ditch LibUnity
In favor of Granite's new functions

Marukesu:
fixes #532


tintou:
Thanks for this, it looks good, if you could just change the Granite version requirements to `>=5.2.5` in the meson file that would be even better :slightly_smiling_face:

Marukesu:
done

PR #599, opened 2021-02-03
Gtk.Dialog → Granite.Dialog, Hdy.Window

danrabbit:
For the editing tools windows, we don't use the action area so there's really no reason for this to be a dialog. It's much easier to make this a Hdy.Window


cassidyjames (changes requested):
I can't drag the handles in the RGB histogram thing without also moving the window, which is jank.

cassidyjames (approved):

PR #603, opened 2021-02-04
CropTool: make properties private set

danrabbit:
Try to reduce interference from outside this class


jeremypw (approved):
LGTM. Crop tool still works as expected.

PR #607, opened 2021-02-06
EditingHostPage: don't track tool windows

danrabbit:
This is jank af


cassidyjames:
yo wtf yeah this is jank

cassidyjames (approved):

PR #609, opened 2021-02-06
PhotoCanvas: Use properties and remove unused

danrabbit:
* Replace getter methods with properties
* Remove unused methods


jeremypw (approved):
Compiles with no extra warnings.  No regressions found in editing tools.

PR #611, opened 2021-02-10
Export Dialog: fix action area margin and cleanup

danrabbit:
Might as well do some code style fixes while we're here


jeremypw (approved):
Works as expected - no regressions found.

I notice that two labels have the same mnemonic letter but that is true in master also.

PR #627, opened 2021-07-24
Restore title to header when viewing single photo

### LINKED ISSUE #626: photo-viewer does not display filename in headerbar
In elementary OS 6 Beta 2 (Odin) when I double-click on a picture in elementary files, the photo-viewer opens the picture. The photo-viewer used to show at least the filename on the headerbar. Now the photo-viewer does not give any information about the picture it is displaying.
photo-viewer-bug

jeremypw:
Fixes #626


danrabbit (changes requested):
I think we need to set both since this also will be used in the multitasking view for the window title

jeremypw:
@danrabbit Well spotted - I have now kept the Gtk.Window title.

I noticed that Library Window uses a more complicated method of keeping the header and window titles in sync but I do not think it is worth it in DirectWindow.

cassidyjames:
Since it appears we're building the same string twice, should we build it once and then set it twice? Or can we just `header.title = title`?

src/direct/DirectWindow.vala
```diff
@@ -80,6 +80,8 @@ public class DirectWindow : AppWindow {
     public void update_title (File file, bool modified) {
         title = "%s%s (%s) - %s".printf ((modified) ? "*" : "", file.get_basename (),
                                          get_display_pathname (file.get_parent ()), _ (Resources.APP_TITLE));
+        header.title = "%s%s (%s) - %s".printf ((modified) ? "*" : "", file.get_basename (),
+                                                get_display_pathname (file.get_parent ()), _ (Resources.APP_TITLE));
```
cassidyjames, line 84: Seems this would work just as well :joy: 

```suggestion
        header.title = title;
```

danrabbit: Yeah this makes sense to me

danrabbit (approved):

danrabbit (approved):

PR #637, opened 2021-08-27
CheckboardLayout: use rubberband widget path to draw selection band

danrabbit:
I'm having some issues here with artifacting on the border. Maybe there's a clear method I'm unaware of? But this makes sure we use actual values for the rubberband element and no hardcoded values

### DIFF
--- a/src/Checkerboard/CheckerboardLayout.vala
+++ b/src/Checkerboard/CheckerboardLayout.vala
@@ -25,9 +25,6 @@ public class CheckerboardLayout : Gtk.DrawingArea {
     // the following are minimums, as the pads and gutters expand to fill up the window width
     public const int COLUMN_GUTTER_PADDING = 24;
 
-    // For a 40% alpha channel
-    private const double SELECTION_ALPHA = 0.40;
-
     // The number of pixels that the scrollbars of Gtk.ScrolledWindows allocate for themselves
     // before their final size is computed. This must be taken into account when computing
     // the width of this widget. This value was 0 in Gtk+ 2.x but is 1 in Gtk+ 3.x. See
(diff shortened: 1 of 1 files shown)


src/Checkerboard/CheckerboardLayout.vala
```diff
@@ -1079,15 +1076,43 @@ public class CheckerboardLayout : Gtk.DrawingArea {
...
+        label_widget_path.iter_set_object_name (-1, "rubberband");
+
+        var rubberband_context = new Gtk.StyleContext ();
+        rubberband_context.set_path (label_widget_path);
+
+        var bg_color = (Gdk.RGBA) rubberband_context.get_property (
+            Gtk.STYLE_PROPERTY_BACKGROUND_COLOR,
+            Gtk.StateFlags.NORMAL
+        );
+
```
jeremypw, line 1093: `StyleContext.get_background_color ()` is deprecated and this seems to do the same thing (although it is not currently deprecated for some reason). You are supposed to use StyleContext.render_background () which should also render the appropriate border and corners.  I'll see whether this will work with this constructed style context.

jeremypw:
It works OK (and is simpler) using `rubberband_context.render_background ()` except there is no frame (unlike the Files selection band).  If I add `rubberband_context.render_frame () then I also get the "echo" of the previous frames :disappointed:   I'll look into the reason for that.

jeremypw:
OK, I've fixed the "echo" - needed to increase the area of the rectangle that was being invalidated by 1px on all sides to ensure the frame is erased.

jeremypw:
@danrabbit Does this look OK to you?

Issue #647, opened 2021-09-30
Segfault when reading exif metadata


asonix:
### What Happened?

There's an image on my computer with the following exif timestamp data:
(captured with imagemagick's `identify -verbose`
```
  Properties:
    date:create: 2019-11-13T09:27:04-06:00
    date:modify: 2018-10-13T00:24:28-05:00
    exif:DateTime: 2018:10:13 24:24:27
```
When Photos attempts to import this image, it crashes (note: shotwell does not crash, and photos on elementary 5.0 did not either).

The backtrace indicates it's failing to parse the exif:DateTime `2018:10:13 24:24:27` into a representable date, and so converting the malformed date to a unix timestamp segfaults.
```
Thread 36 "pool-io.element" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffb57fa700 (LWP 57332)]
0x00007ffff7ebc668 in g_date_time_to_unix () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
(gdb) backtrace
#0  0x00007ffff7ebc668 in g_date_time_to_unix () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x0000555555678c29 in metadata_date_time_from_exif_date_time (date_time_s=0x7fff98324160 "2018:10:13 24:24:27", timestamp=0x7fffb57f8f38) at ../src/MediaMetadata.vala:121
#2  0x0000555555678417 in metadata_date_time_construct_from_exifPython Exception <class 'ValueError'> Variable 'static_fundamental_type_nodes' not found.: 
 (object_type=, label=0x7fff98324160 "2018:10:13 24:24:27", error=0x7fffb57f8fb0)
    at ../src/MediaMetadata.vala:68
#3  0x0000555555678554 in metadata_date_time_new_from_exif (label=0x7fff98324160 "2018:10:13 24:24:27", error=0x7fffb57f8fb0) at ../src/MediaMetadata.vala:67
#4  0x000055555581dcf2 in photo_metadata_get_date_time (self=0x55555b1cbb60, tag=0x7fff9831e1b0 "Exif.Image.DateTime") at ../src/photos/PhotoMetadata.vala:568
#5  0x000055555581df08 in photo_metadata_get_first_date_time (self=0x55555b1cbb60, tags=0x7fffbc00e370, tags_length1=6) at ../src/photos/PhotoMetadata.vala:579
[...]

PR #648, opened 2021-09-30
Fix invalid date segfault

asonix:
This fixes date parsing enough to not segfault anymore for my own photo collection.

Fixes https://github.com/elementary/photos/issues/647

### DIFF
--- a/src/MediaMetadata.vala
+++ b/src/MediaMetadata.vala
@@ -117,6 +117,34 @@ public class MetadataDateTime {
             return false;
         }
 
+        if (month > 12 || hour >= 24 || minute >= 60 || second >= 60) {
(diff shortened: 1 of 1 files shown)


jeremypw:
Thanks for the fix!  Maybe you can use `GLib.DateYear` to check whether it is a leap year?

tintou:
What you just need here is the null-check:
```vala
GLib.DateTime? date_time = new DateTime.local (year, month, day, hour, minute, (double) second);
if (date_time == null) {
    return false;
}

timestamp = date_time.to_unix ();
return true;
```

davidmhewitt (approved):
Thanks for the fix!

PR #651, opened 2021-10-01
Fix slow reordering.

### LINKED ISSUE #649: Very slow to open due to reorder of massive tree
### What Happened?

On opening photos, it kicks off a `Branch.reorder_all` (https://github.com/elementary/photos/blob/master/src/sidebar/Branch.vala#L337), which takes on the order of minutes on my computer to complete. This makes Photos effectively unusable unless I go to make coffee between launching it and starting to use it.

At first, I thought photos was stuck, since it sat consuming 100% CPU on one core for so long and elementary OS launched the "Photos isn't responding" window, but after letting it sit, it eventually finishes.

### Steps to Reproduce

1. Have a very large collection of photos (86GB) neatly organized into folders based on year/month/day/photo
2. Open photos

### Expected Behavior

It should open in a reasonable amount of time

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

```shell
There's no logs in the path that's taking the time
```

### Hardware Info

```
Linux firestar 5.11.0-37-generic #41~20.04.2-Ubuntu SMP Fri Sep 24 09:06:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

01:00.0 VGA compatible controller: NVIDIA Corporation GP104 [GeForce GTX 1070] (rev a1)

nvidia-driver-470/focal,now 470.74-0ubuntu0.20.04.1 amd64 [installed]
```

vjr:
Fixes #649

### DIFF
--- a/src/sidebar/Branch.vala
+++ b/src/sidebar/Branch.vala
@@ -170,7 +170,7 @@ public class Sidebar.Branch : Object {
         }
 
         public void reorder_children (bool recursive, ChildrenReorderedCallback cb) {
-            if (children == null)
+            if (children == null || children.size <= 1)
                 return;
 
             Gee.SortedSet<Node> reordered = new Gee.TreeSet<Node> (comparator_wrapper);


vjr:
This is only a potential fix because I've not tested it thoroughly.

asonix:
This doesn't seem to change the time it takes for the reorder to complete on my laptop. I tried comparing this branch to master and didn't notice a difference

vjr:
OK, I'll try looking into this some more...

PR #654, opened 2021-10-19
Group day events

### LINKED ISSUE #628: Group multiple events for 1 date
## What Happened
All pictures are added in seperate events.

## Expected Behavior
Pictures from the same date should be grouped in 1 event

## Steps to Reproduce

1.Add photo library
2. Open folder
3. Look at multiple events for 1 date

## Platform Information

Elementary OS 6.0 Beta 2
Photos app version 2.7.1
Schermafdruk van 2021-07-30 15-16-42

jeremypw:
Fixes #628

### DIFF
--- a/src/Views/EventsDirectoryPage.vala
+++ b/src/Views/EventsDirectoryPage.vala
@@ -321,14 +321,18 @@ public class SubEventsDirectoryPage : EventsDirectoryPage {
     public enum DirectoryType {
         YEAR,
         MONTH,
+        DAY,
         UNDATED;
     }
 
     public const string UNDATED_PAGE_NAME = _ ("Undated");
     public const string YEAR_FORMAT = _ ("%Y");
     public const string MONTH_FORMAT = _ ("%OB");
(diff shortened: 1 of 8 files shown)


jeremypw:
PhotosSidebarGroupByDay

marco4mp:
I agree with @stan-janssen. By default it should be have three levels of hierarchy:

1.... Year
2......... Month
3.............. Events (Photos grouped automatically according to the day they were taken)


Captura realizada el 2021-11-18 20 02 43

jeremypw:
@marco4mp So the third level of hierarchy would open a view containing multiple events (photos) taken on that day rather than expanding to show a fourth level?  That should be doable.

marco4mp:
I think that Events are something like Albums.  
Photos makes the mistake of taking each photo as an event (or album) in itself.

So the expected behavior would be that If I click on one of the events in the side panel, this should show me all the photos that were taken on that same day. As shown in the following example image:
Photos_3_Event

jeremypw:
Closing in favour of #669 which addresses the root cause of the incorrect Event id allocation on importation.

PR #666, opened 2021-11-12
Simplify StraightenToolWindow class

ZenitsuDev:
StraightenToolWindow has a base function which derives from EditingToolWindow, although the constructor of EditingToolWindow only sets transient_for property to the container, so transient_for can be called directly from StraightenToolWindow constructor itself without depending on the Parent Class (EditingToolWindow). The other functions included in the constructor are moved to the construct block to simplify further. Thanks @danrabbit

### DIFF
--- a/src/editing_tools/StraightenTool.vala
+++ b/src/editing_tools/StraightenTool.vala
@@ -121,19 +121,19 @@ public class StraightenTool : EditingTool {
         public StraightenToolWindow (Gtk.Window container) {
             Object (transient_for: container);
         }
-        
+
         construct {
             angle_slider.set_size_request (MIN_SLIDER_SIZE, -1);
             angle_slider.set_value (0.0);
             angle_slider.set_draw_value (false);
-            
+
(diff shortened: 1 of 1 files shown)


danrabbit:
@ZenitsuDev this looks good purely from a functional perspective! Nice simple fix.

The linter is complaining because there's some spaces introduced on line 124. Since this is a blank lane, it should be completely blank

danrabbit (changes requested):
See above :point_up:

ZenitsuDev:
Greetings. I changed it. It should be fixed now :)

danrabbit:
@ZenitsuDev if you look at the "files changed" tab, you can see that you've now introduced a bunch of spaces on a bunch of other lines somehow :sweat_smile:

ZenitsuDev:
I'm sorry about that :) It should be fixed by now :)

danrabbit (approved):
Hey there we go. Looks good. Nice job!

ZenitsuDev:
thanks @danrabbit !

PR #668, opened 2021-11-19
Make Command Class Gobject-style Constructed

ZenitsuDev:
The Command class which is the parent class of the editing tools can be converted into gobject-style construction, which will also enable editing_tools to be gobject-style constructed.

### DIFF
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -339,7 +339,7 @@ public abstract class AppWindow : PageWindow {
         }
 
         if (desc != null) {
-            button.tooltip_text = "%s %s".printf (default_explanation, desc.get_name ());
+            button.tooltip_text = "%s %s".printf (default_explanation, desc.name);
             action.set_enabled (true);
         } else {
             button.tooltip_text = default_explanation;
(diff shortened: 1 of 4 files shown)


jeremypw (changes requested):
Not sure that a construct property should be virtual as it is immutable and the getter should just return it as is (?)  Otherwise it would be better to leave the virtual functions and rename them to something like `get_display_name ()` to show that they do not necessarily return the actual name property.  Could you check the usage of these functions in the rest of the code and decide the appropriate course?

jeremypw:
Also, I think that the property getter will return an unowned string whereas the explicit function will return a string copy.

ZenitsuDev:
okay I will try fixing it

jeremypw:
On closer examination of Photos code it seems that the CommandDescription interface is over-engineering (like a lot of the code).  It is only used in one function!  So I have removed it and replaced it with direct access to Command properties.

ZenitsuDev:
okay. Then

ZenitsuDev:
oh. It is something I was not able to think of right away. Thanks anyway for the new knowledge

jeremypw:
I have pushed a commit to this branch removing that interface.

For future reference, it is better to create a new branch on your fork and change that rather than changing the master branch.

ZenitsuDev:
Okay. Thank You!

ZenitsuDev (commented):
I think it is good! Thanks for the help

jeremypw (approved):

PR #675, opened 2021-11-25
Fix invalid cast  from Video to Photo

jeremypw:
This fixes a terminal warning about casting a Video to a Photo when Photos is opened and there are Videos in the library.

Most of the checks before casting are probably not actually necessary but were fixed anyway.

Also the second commit fixes a problem where the context menu on Videos had a "Open in" option but the submenu was empty.

 Video handling in Photos (e.g. thumbnailing) is not very good.  It needs to be decided whether it is worth continuing video support before spending the time to improve it.

### DIFF
--- a/src/Commands.vala
+++ b/src/Commands.vala
@@ -1199,7 +1199,11 @@ public class SetRawDeveloperCommand : MultipleDataSourceCommand {
         last_transformation_map = new Gee.HashMap<Photo, PhotoTransformationState> ();
 
         foreach (DataView view in iter) {
-            Photo? photo = view.source as Photo;
+            Photo? photo = null;
+            if (view.source is Photo) {
+                photo = view.source as Photo;
+            }
+
             if (is_raw_photo (photo)) {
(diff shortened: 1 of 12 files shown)


src/Commands.vala
```diff
@@ -1226,7 +1234,11 @@ public class SetRawDeveloperCommand : MultipleDataSourceCommand {
     }
 
     public override void undo_on_source (DataSource source) {
-        Photo? photo = source as Photo;
+        Photo? photo = null;
+        if (source is Photo) {
+            photo = source as Photo;
+        }
```
danrabbit, line 1240: Isn't this what `as` already does as compared to an unsafe cast? It'll return `null` if the type doesn't match?

jeremypw: Yes, but it is also giving those terminal criticals.  I can rewrite using unsafe casts and using `source is Photo` instead of `photo != null` if you like.

jeremypw:
Leaving as draft until other component fixes are merged. See #689, #690, #691

jeremypw:
Closing as out of date.

PR #676, opened 2021-11-26
Make AdjustTool Class Gobject-style constructed

ZenitsuDev:
Revises the AdjustTool class to follow gobject-style construction and the recommended code style.


src/editing_tools/AdjustTool.vala
```diff
@@ -371,7 +419,8 @@ public class EditingTools.AdjustTool : EditingTool {
 
         /* set up temperature & tint */
         TemperatureTransformation temp_trans = (TemperatureTransformation)
```
ryonakano, line 421: Or we can use `var` for the left operand, like this:

```vala
var temp_trans = (TemperatureTransformation) transformations.get_transformation (
    PixelTransformationType.TEMPERATURE
);
```

src/editing_tools/AdjustTool.vala
```diff
@@ -418,10 +470,11 @@ public class EditingTools.AdjustTool : EditingTool {
         virgin_histogram_pixbuf = histogram_pixbuf.copy ();
 
         DataCollection? owner = canvas.photo.get_membership ();
-        if (owner != null)
+        if (owner != null) {
             owner.items_altered.connect (on_photos_altered);
+        }
 
-        base.activate (canvas);
+        activate (canvas);
```
ryonakano, line 477: Hmm isn't the `base.` here necessary? This should call the original activate method of the base class

src/editing_tools/AdjustTool.vala
```diff
@@ -447,16 +501,17 @@ public class EditingTools.AdjustTool : EditingTool {
         draw_to_pixbuf = null;
         fp_pixel_cache = null;
 
-        base.deactivate ();
+        deactivate ();
```
ryonakano, line 504: Here too

ryonakano (changes requested):
Thank you for your contribution! Leaving some comments.

src/editing_tools/AdjustTool.vala
ZenitsuDev: I think so too :) Thanks!

src/editing_tools/AdjustTool.vala
ZenitsuDev: Thinking about it, yeah, probably, or the call would have been this.activate

ZenitsuDev:
Konnichiwa @ryonakano -san, thanks for the comment! I'll be sure to improve this

ryonakano (approved):
Looks good, thank you!

PR #680, opened 2021-11-26
Replace all occurance of 'Shotwell' to 'Photos' in the UI

ryonakano:
Replace all occurance of 'Shotwell' to 'Photos' where users can see.


tintou (approved):

hugok79:
👍

PR #683, opened 2021-12-01
Make RedeyeTool Class Gobject-style constructed

ZenitsuDev:
Make redeyetool gobject-style constructed. Also replaced specified type with `var` when there is clear indication of the instance.

### DIFF
--- a/src/editing_tools/RedEyeTool.vala
+++ b/src/editing_tools/RedEyeTool.vala
@@ -35,7 +35,7 @@ public struct EditingTools.RedeyeInstance {
     }
 
     public static Gdk.Rectangle to_bounds_rect (EditingTools.RedeyeInstance inst) {
-        Gdk.Rectangle result = Gdk.Rectangle ();
+        var result = Gdk.Rectangle ();
         result.x = inst.center.x - inst.radius;
         result.y = inst.center.y - inst.radius;
         result.width = 2 * inst.radius;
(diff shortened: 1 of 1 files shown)


src/editing_tools/RedEyeTool.vala
```diff
@@ -70,8 +70,11 @@ public class EditingTools.RedeyeTool : EditingTool {
                 RedeyeInstance.MIN_RADIUS, RedeyeInstance.MAX_RADIUS, 1.0);
 
         public RedeyeToolWindow (Gtk.Window container) {
-            base (container);
+            Object (transient_for: container);
```
ryonakano, line 74: Can you remove this extra new line while we're here?

ZenitsuDev: Oh I did not notice. I will. Thanks

src/editing_tools/RedEyeTool.vala
```diff
@@ -45,9 +45,9 @@ public struct EditingTools.RedeyeInstance {
     }
 
     public static RedeyeInstance from_bounds_rect (Gdk.Rectangle rect) {
-        Gdk.Rectangle in_rect = rect;
+        var in_rect = rect;
```
ryonakano, line 48: Hmm, it looks like we may actually not need the variable `in_rect`; it's essentially the same with the argument `rect` and only an  unnecessary copy happens heere. Maybe we can remove this and replace the occurance of `in_rect` to `rect` in  this method while we're here.

ryonakano (commented):
Just one more nitpicking comment, otherwise looking good.

ZenitsuDev:
I'll try that. thanks

ZenitsuDev:
Should be fine by now. Thanks again @ryonakano

ryonakano (approved):
Looks good, thank you!

PR #686, opened 2021-12-01
GObject-construct PhotoCanvas

ZenitsuDev:
GObject-construct PhotoCanvas.vala, the last one inside src/editing_tools. Also fixed some code style conflicts.

### DIFF
--- a/src/editing_tools/PhotoCanvas.vala
+++ b/src/editing_tools/PhotoCanvas.vala
@@ -37,9 +37,9 @@ public abstract class EditingTools.PhotoCanvas : Object {
     public Dimensions surface_dim { get; set construct; }
     public Gdk.Pixbuf scaled_pixbuf { get; set construct; }
     public Gdk.Rectangle scaled_position { get; set construct; }
-    public Gdk.Window drawing_window { get; set construct; }
-    public Gtk.Window container { get; set construct; }
-    public Photo photo { get; set construct; }
+    public Gdk.Window drawing_window { get; construct; }
+    public Gtk.Window container { get; construct; }
+    public Photo photo { get; construct; }
 
     private Cairo.Surface scaled { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/editing_tools/PhotoCanvas.vala
```diff
@@ -29,30 +29,35 @@
...
-    public Gdk.Window drawing_window { get; private set; }
-    public Gtk.Window container { get; private set; }
-    public Photo photo { get; private set; }
+    public Cairo.Context default_ctx { get; set construct; }
+    public Dimensions surface_dim { get; set construct; }
+    public Gdk.Pixbuf scaled_pixbuf { get; set construct; }
+    public Gdk.Rectangle scaled_position { get; set construct; }
+    public Gdk.Window drawing_window { get; set construct; }
+    public Gtk.Window container { get; set construct; }
+    public Photo photo { get; set construct; }
```
ryonakano, line 42:
```suggestion
    public Gdk.Window drawing_window { get; construct; }
    public Gtk.Window container { get; construct; }
    public Photo photo { get; construct; }
```

These properties seems to be substituted only in the constructor, so I guess we can remove the `set` accels here

src/editing_tools/PhotoCanvas.vala
```diff
@@ -29,30 +29,35 @@
...
 
-    private Cairo.Surface scaled;
+    private Cairo.Surface scaled { get; set; }
```
ryonakano, line 44:
```suggestion
    private Cairo.Surface scaled;
```

We don't need to explicit `{ get; set; }` here, this is a private member so it's guaranteed that it can't be accessed out of the class

ryonakano (commented):
Two nitpicking comments.

ZenitsuDev:
Understood well. Thanks!

ryonakano (approved):
Nice work! :+1:

ZenitsuDev:
Thanks again @ryonakano !

PR #688, opened 2021-12-01
Fix Minor Code-style Conflicts

ZenitsuDev:
Fixes very minor code-style conflicts I overlooked when I made a pr before. (Does not change any code functionally)


src/editing_tools/AdjustTool.vala
```diff
@@ -419,28 +419,24 @@ public class EditingTools.AdjustTool : EditingTool {
 
         /* set up temperature & tint */
         var temp_trans = (TemperatureTransformation) transformations.get_transformation (
-            PixelTransformationType.TEMPERATURE
-        );
+            PixelTransformationType.TEMPERATURE);
```
ryonakano, line 422: Hmm, I don't think we need to change this :thinking:

src/editing_tools/AdjustTool.vala
```diff
@@ -134,6 +132,7 @@ public class EditingTools.AdjustTool : EditingTool {
             var button_box = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL) {
                 margin_top = 12
             };
+
```
ryonakano, line 135: Nitpicking but this has a trailing `;` we don't put a newline here IMO.

Have a newline:

```
if (i < 10) {
    text = "";
}

var grid = new Gtk.Grid ();
…
```

No newline:

```
var grid = new Gtk.Grid () {
    margin = 6
};
grid.add (label);
…
```

src/editing_tools/AdjustTool.vala
ZenitsuDev: Oh okay. Thanks

ZenitsuDev:
Should be fine now. Thanks @ryonakano

src/editing_tools/AdjustTool.vala
```diff
@@ -142,6 +140,7 @@ public class EditingTools.AdjustTool : EditingTool {
                 column_spacing = 12,
                 row_spacing = 12
             };
+
```
ryonakano, line 143:
```suggestion
```

src/editing_tools/AdjustTool.vala
```diff
@@ -767,49 +767,49 @@ public class EditingTools.AdjustTool : EditingTool {
...
+            case PixelTransformationType.TONE_EXPANSION:
+                var expansion = (ExpansionTransformation) transformation;
+
+                if (!disable_histogram_refresh) {
+                    adjust_tool_window.histogram_manipulator.set_left_nub_position (
+                        expansion.get_black_point ());
+                    adjust_tool_window.histogram_manipulator.set_right_nub_position (
+                        expansion.get_white_point ());
+                }
+                break;
```
ryonakano, line 779:
```suggestion
                }

                break;
```

ryonakano (commented):
Two tiny things and ready to go

ZenitsuDev:
Should be okay now

src/editing_tools/AdjustTool.vala
```diff
@@ -216,7 +214,7 @@ public class EditingTools.AdjustTool : EditingTool {
         }
 
         public override bool compress (Command command) {
-            AdjustResetCommand reset_command = command as AdjustResetCommand;
+            var reset_command = (AdjustResetCommand) command;
             if (reset_command == null) {
                 return false;
             }
```
ryonakano, line 220: Oh wait, if we change the way of casting from `as AdjustResetCommand` to `(AdjustResetCommand)`, the `if` sentence following from the next lines might no longer work? Casting by `as` can return null but the latter don't and maybe throws an error I guess, as I read [the wiki](https://wiki.gnome.org/Projects/Vala/Tutorial#:~:text=gobject%2Dstyle%20construction.-,Dynamic%20Type%20Casting,-For%20the%20dynamic) (sorry if I'm misreading).

ZenitsuDev: I read about it too on the code style, it says it throws a null. but since the code style says we should avoid `as`, I guess we can just remove the if statement which follows.

ZenitsuDev:
@ryonakano I also removed the one at line ~345

src/editing_tools/AdjustTool.vala
```diff
@@ -216,10 +214,7 @@ public class EditingTools.AdjustTool : EditingTool {
         }
 
         public override bool compress (Command command) {
-            AdjustResetCommand reset_command = command as AdjustResetCommand;
-            if (reset_command == null) {
-                return false;
-            }
+            var reset_command = (AdjustResetCommand) command;
```
jeremypw, line 217: Can it be ensured that `command` is not null and is of type `AdjustResetCommand`?
A safer ways is:
```suggestion
            if (command == null || !(command is AdjustResetCommand)) {
                return false;
            }
            
            var reset_command = (AdjustResetCommand) command;
```

Or maybe instead add a `requires` clause to the function signature if there is no legitimate reason for `command` not to be of the expected type.

jeremypw (commented):
Just a comment about lack of casting safety checks and two possible alternatives.

src/editing_tools/AdjustTool.vala
ZenitsuDev: I think this is enough. Or should I use a requires clause instead?

ZenitsuDev:
@jeremypw I think the if statement for verification that it will not be null is enough. Or should I use a requires clause instead?

jeremypw:
I am not entirely sure what happens when the `requires` clause is failed or what the likelihood of that happening is.  So better to leave the explicit handling at the moment.

src/editing_tools/AdjustTool.vala
```diff
@@ -344,10 +343,7 @@ public class EditingTools.AdjustTool : EditingTool {
                 return photo.equals (owner.canvas.photo);
             }
 
-            AdjustEnhanceCommand enhance_command = (AdjustEnhanceCommand) command;
-            if (enhance_command == null) {
-                return false;
-            }
+            var enhance_command = (AdjustEnhanceCommand) command;
```
jeremypw, line 346: Need safety check

src/editing_tools/AdjustTool.vala
```diff
@@ -532,7 +529,7 @@ public class EditingTools.AdjustTool : EditingTool {
     }
 
     private void on_reset () {
-        AdjustResetCommand command = new AdjustResetCommand (this, transformations);
+        var command = new AdjustResetCommand (this, transformations);
```
jeremypw, line 532: Need safety check

jeremypw:
There is another place where a check is better.  I have assumed that in the switch statement no additional checks are required because the case statement already ensures the type.

src/editing_tools/AdjustTool.vala
ZenitsuDev: I see. I'll probably edit the others I see along the way

ZenitsuDev:
> There is another place where a check is better. I have assumed that in the switch statement no additional checks are required because the case statement already ensures the type.

I'll look at it now and try to find it.

src/editing_tools/AdjustTool.vala
jeremypw:
```suggestion
            if (command == null || !(command is AdjustEnhanceCommand) {
                 return false;
            }
     
            var enhance_command = (AdjustEnhanceCommand) command;
```

There is another place above this where a similar check should be used for safety but it has not yet been edited so I cannot make a suggestion.

jeremypw:
@ZenitsuDev Nearly there! just need fixing the lint - there is some trailing white-space somewhere. You can pre-check a commit for lint by installing and using https://github.com/vala-lang/vala-lint.

ZenitsuDev:
Sorry @jeremypw I wasn't able to reply right away. btw, I fixed it now. It should be fine now. Thanks

src/editing_tools/AdjustTool.vala
```diff
@@ -342,13 +342,16 @@ public class EditingTools.AdjustTool : EditingTool {
                 // multiple successive enhances are as good as a single, as long as it's on the
                 // same photo
                 return photo.equals (owner.canvas.photo);
+            } else {
+                return false;
```
jeremypw, line 346: This makes the following code unreachable.  I do not think this change should be made.

jeremypw (changes requested):
Just noticed a new terminal warning on compiling.  A change you made makes some code unreachable.

src/editing_tools/AdjustTool.vala
ZenitsuDev: Yeah I agree. Lemme change that

ZenitsuDev:
@jeremypw I removed the subsequent if block I added since if the command is null, it'll go to the else anyway, and the type check I don't think it is necessary since the casting was did prior to the if statement. Sorry for not looking at these critically.

jeremypw:
@ZenitsuDev I am afraid there is still an unreachable code warning on compiling this PR.  I have submitted a pull request to this branch on your fork fixing the issue (I do cannot modify this PR directly). 
While fixing the unreachable code, I removed some null checks that I realised were unnecessary as the function signatures do not allow the functions to run with a null parameter.  I checked that if that happens the app does not crash - you just get a critical warning in the terminal.  I think it unlikely that these functions will get called with null in reality.
I also changed some remaining casts of `command` to the recommended format.

ZenitsuDev:
Okay I will merge it. thanks again @jeremypw

jeremypw (approved):
OK., all the adjustment tools seem to work as normal and undo works as well.  I think this is good to go.  Thanks!

ZenitsuDev:
yay! Thanks @jeremypw

PR #689, opened 2021-12-03
Check video date valid on importing

jeremypw:
Photos was found to crash on importing certain video files - the metadata was found to have a creation date not in the expected format.  As a quick fix, any metadata creation date not immediately convertible to a valid GLib.DateTime is rejected as a timestamp and the current date used so that importation can succeed.  Parsing other possible date formats is left for a future PR.


zeebok (approved):
I don't have a sample video to work with to test this but the code looks reasonable!

PR #723, opened 2022-08-29
Fix empty Open with context submenus

jeremypw:
Extracted from #675 to reduce diff.

When running as a Flatpak the "Open with" submenus currently return no items but still appear as an empty menu.  This issue occurs on the Collection, Library and Direct pages. This PR fixes that.

This is effectively the same code repeated three times in different classes as virtually the same code appears 3 times in master.

DRYing is therefore possible but is left for a future PR (unless requested here).  Most of the extra lines are due to code style improvements.  The only change in logic is to return the number of items added to the submenu and set the sensitivity of the corresponding menu-item accordingly.

### DIFF
--- a/src/Views/CollectionPage.vala
+++ b/src/Views/CollectionPage.vala
@@ -253,19 +253,32 @@ public abstract class CollectionPage : MediaPage {
             item_context_menu.add (remove_menu_item);
             item_context_menu.add (trash_menu_item);
             item_context_menu.show_all ();
-        }
 
-        Photo? photo = (get_view ().get_selected_at (0).source as Photo);
-        if (photo != null) {
-            unowned PhotoFileFormat photo_file_format = photo.get_master_file_format ();
-            populate_external_app_menu (open_menu, photo_file_format, false);
+             int n_items = 0;
+             var source = get_view ().get_selected_at (0).source;
+             if (source != null && source is Photo) {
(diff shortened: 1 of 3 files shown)


src/Views/CollectionPage.vala
```diff
@@ -324,24 +337,20 @@ public abstract class CollectionPage : MediaPage {
...
-
         if (!raw) {
             var files_appinfo = AppInfo.get_default_for_type ("inode/directory", true);
-
-            var files_item_icon = new Gtk.Image.from_gicon (files_appinfo.get_icon (), Gtk.IconSize.MENU);
-            files_item_icon.pixel_size = 16;
-
-            var menuitem_grid = new Gtk.Grid ();
+            var files_item_icon = new Gtk.Image.from_gicon (
+                files_appinfo.get_icon (), Gtk.IconSize.MENU
```
danrabbit, line 351: When we get app icons from appinfo we need to specify pixel size or Gtk won't resize images, it'll just use whatever the nearest size is. So for example if a file browser were to only supply a 128px icon, that's what Gtk would use at 128px instead of scaling it down to 16px

jeremypw: Done

zeebok (approved):
I know you requested Dani to look but this looks good to me as well, if we want to move forward getting this merged in

PR #732, opened 2023-01-23
Fix translation charset

meisenzahl:
Right now we are not able to build Photos on Fedora 37:

https://github.com/meisenzahl/distro-agnostic/actions/runs/3987907130/jobs/6838343834#step:5:16634

This small change fixes this:

https://github.com/meisenzahl/distro-agnostic/actions/runs/3988175247/jobs/6838977411#step:5:18232

@ryonakano do you have any concerns changing the charset?


tintou (approved):
We are forcing UTF-8 from the [code-side](https://github.com/elementary/photos/blob/master/src/main.vala#L369) anyway so better use it here too

PR #746, opened 2023-08-21
Remove the publishing plugins

### LINKED ISSUE #718: support for webkit2gtk-4.1 / rest-1 / libsoup-3
Starting with GNOME 43, some core components of the desktop have been ported to libsoup-3, which apparently brings a lot of new features, including HTTP/2 support. However, applications and libraries **cannot** load both libsoup-2.4 and libsoup-3 in the same process, as this is unsupported and will crash the application immediately.

elementary Photos will also be affected this change eventually, but at least not immediately. It links against libsoup-2.4, rest-0.7, and webkit2gtk-4.0, but since those three libraries will remain available and compatible with each other, at least it will continue to build and work on GNOME 43. However, webkit2gtk-4.0 will be removed by upstream GNOME in the near-to-mid-future, so elementary Photos needs to support libsoup-3 and to webkit2gtk-4.1 sooner rather than later.

Additionally, if one of the other GNOME libraries that's linked to by elementary Photos gets ported from libsoup-2.4 to libsoup-3, Photos will also be in trouble even earlier than the webkit2gtk-4.0 removal date.

---

EDIT: Also mention rest-0.7, which is only compatible with libsoup-2.4, and rest-1, which is only compatible with libsoup-3.

### LINKED ISSUE #742: Evaluate the publishing plugins
Are they still working?

In pantheon-photos-publishing:
 * [ ] FacebookPublishing
 * [ ] PiwigoPublishing
 * [ ] FlickrPublishing
 * [ ] PicasaPublishing
 * [ ] YouTubePublishing

In pantheon-photos-publishing-extras:
 * [ ] TumblrPublishing
 * [ ] YandexPublishing

tintou:
Closes #742

Closes #718

Please use "Rebase and merge"


danirabbit (approved):
Let the past die

PR #751, opened 2023-10-15
POTFILES: Remove the publishing plugins

ryonakano:
These files are removed in #746 and causing the gettext actions failing


davidmhewitt (approved):
Thank you!

PR #762, opened 2024-06-06
Use open directory portal

danirabbit:
We can't read other app's appinfo in the sandbox, so remove the open in submenu. There is a portal for showing a file in Files, so use that portal


zeebok:
I think my only real reservation is that this creates a tight dependency on Files being on the user's system

danirabbit:
@zeebok Since it uses the portal it launches whatever file browser you have installed :)

zeebok (approved):
A lot more repeated code than I would've expected

danirabbit:
Yeah this project is wild in that it has so much abstraction and none of it is DRY lol

PR #765, opened 2024-06-10
Rewrite packaging to pull in Flatpak

danirabbit:
Rewrites packaging to pull in the Photos Flatpak release


danirabbit:
Turns out the photo viewer doesn't work in Flatpak. Not sure if that's easy to fix or not

zeebok (approved):
This is a little out of my wheelhouse but what I see seems reasonable and makes sense

PR #774, opened 2024-07-24
Flatpak: add sandbox hole for home

danirabbit:
It turns out we need a sandbox hole for home because the photos viewer can't open files outside the sandbox


zeebok (approved):
I was going to test real quick just to be safe but I am getting 403 errors when trying to build the flatpak (coming from gnome's repo) but this seems good and makes sense based on the convo in the issue

PR #778, opened 2024-09-18
Modernize desktop files, simplify build

danirabbit:
* Don't set exec name using build system
* Add `SingleMainWindow` key
* Remove unused `GenericName` key


jeremypw (approved):
LGTM

PR #781, opened 2024-11-21
Make Photometadata class more robust

jeremypw:
May fix some issues with loading corrupt metadata.

 - Replaces an assert statement with throwing errors
 - Replaces deprecated GExiv2 methods with new error throwing ones and handle errors where necessary

### DIFF
--- a/src/MediaMetadata.vala
+++ b/src/MediaMetadata.vala
@@ -44,7 +44,7 @@ public struct MetadataRational {
     }
 
     public bool is_valid () {
-        return (is_component_valid (numerator) && is_component_valid (denominator));
+        return (denominator > 0 && is_component_valid (numerator) && is_component_valid (denominator));
     }
 
     public string to_string () {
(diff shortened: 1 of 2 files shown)


jeremypw:
The probable cause for reported crashes has now been found and fixed (see #785 ) so the changes here are mainly cosmetic - there is no evidence at the moment that they give rise to crashes.  As there are no test defined in the codebase any regressions will have to be found by user testing I guess :disappointed:  The codebase is littered with myriad functions that return null but in practice seems fairly stable so whether it is worth the work of updating it is debatable.

jeremypw:
If desired I can separate smaller PRs for individual fixes that seem more worthwhile than others.

zeebok (approved):
This code seems good to me. Don't have a file to test with but *shrug*

jeremypw:
I don't think any change here will alter behaviour - just create more error messages so there *shouldn't* be any regressions. I have tested it on my own collection without problems.  I found a set of test images for PNG format at https://github.com/YahooArchive/pngjs-image/tree/master/test/png/PngSuite and these were imported without crashing (some .raw images were rejected).

zeebok:
Yeah, I have a whole butt load of images from taking photos and the it doesn't crash but it wasn't crashing before either so I can't test if it crashes less :P

PR #782, opened 2024-11-22
Make MetadataWriter more robust

jeremypw:
- Add some null checks
 - Remove asserts and replace with critical messages

The codebase contains a *lot* of places where the app just terminates due to an assertion failure.  This does not seem necessary for production code and these have been replaced with critical messages and the subsequent code either omitted or continued as seems appropriate.  If reviewers feel an assertion failure would cause user data corruption on continued use then it should be replaced with a call to AppWindow.panic which at least logs a message before terminating.

There were a few places where a null could theoretically be dereferences. It is difficult to know whether this occurs in practice as the code is extremely convoluted.

### DIFF
--- a/src/MetadataWriter.vala
+++ b/src/MetadataWriter.vala
@@ -521,19 +521,21 @@ public class MetadataWriter : Object {
 
         // mark all the photos as dirty
         if (!already_marked) {
-            try {
-                LibraryPhoto.global.transaction_controller.begin ();
-
-                foreach (LibraryPhoto photo in photos)
+            // Ensure transaction committed by moving outside try/catch
+            LibraryPhoto.global.transaction_controller.begin ();
+            foreach (LibraryPhoto photo in photos) {
+                try {
                     photo.set_master_metadata_dirty (true);
-
-                LibraryPhoto.global.transaction_controller.commit ();
-            } catch (Error err) {
-                if (err is DatabaseError)
-                    AppWindow.database_error ((DatabaseError) err);
-                else
-                    error ("Unable to mark metadata as dirty: %s", err.message);
+                } catch (Error err) {
+                    if (err is DatabaseError) {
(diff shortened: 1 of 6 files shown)


stsdc:
At this point I'm expecting you also to approve this PR :smile: 

obraz

jeremypw:
On further consideration it is not necessary to explicitly null check in the `blacklist_file` functions - the Vala compiler automatically inserts a null check.  One disadvantage of relying on this is that is not immediately obvious what is returned if the parameter requirements are not met when the function signature returns something.  In fact it seems that boolean functions return `false` (safe enough) and functions returning a pointer return `null` (might be unsafe).

src/MetadataWriter.vala
```diff
@@ -58,38 +58,22 @@ public class MetadataWriter : Object {
...
-                    photo.persist_editable_metadata (metadata, out reimport_editable_state);
-                } finally {
-                    LibraryMonitor.unblacklist_file (photo.get_editable_file ());
-                }
+                // Photo does necessary checks and may throw error
+                photo.persist_editable_metadata (metadata, out reimport_editable_state);
             }
         }
 
+        // Vala compiler returns false if metadata null
```
zeebok, line 76: Does Vala do this check because the argument isn't ending with `?` or something else?

src/MetadataWriter.vala
```diff
@@ -588,9 +581,10 @@ public class MetadataWriter : Object {
 
         if (dirty.contains (photo)) {
             bool removed = dirty.remove_first (photo);
-            assert (removed);
-
-            assert (!dirty.contains (photo));
+            if (!removed || dirty.contains (photo)) {
```
zeebok, line 584: Checking dirty seems redundant here since it is within an if-statement of the same check

zeebok:
I just have a couple comments, and there's a conflict to fix. Otherwise this seems good to me

PR #783, opened 2024-11-22
Make BatchImport more robust

jeremypw:
In order to try to get my head round what is going on in BatchImport I ended up splitting out all 18 or so classes/enums contained within that single file (!) and putting then in a BatchImport folder.  I have used slightly more informative filenames but have not (yet) changed the names of the entities themselves although they can be confusing.

I also moved one class (CameraImportJob) out of another mega file "ImportPage.vala" as it was derived from an abstract class in BatchImport.vala. Probably other classes in there that should be split out and put under their own folder.

The diff looks scary but I have changed very little code (see individual commits). Most of the added lines are headers.


jeremypw:
I am going to close this as I doubt there is much appetite for making extensive changes to the codebase without a good reason.  The cause of the crashes on import that were reported have hopefully been fixed by #785

PR #784, opened 2024-11-23
Ignore sniffed jobs below minimum size

jeremypw:
Extracted from #783.  Silences some terminal warnings when folder being imported from contains files named like an image or video but containing little or no data.

What would be a suitable minimum size for Photos to process?


zeebok (approved):
How did you land on 64? I assume this is bytes?

PR #791, opened 2024-12-05
Fix some terminal warnings

jeremypw:
- [x] Test MediaSource before casting to Photo (may be Video)
 - [x] Fix context menu warning if Video item
 - [x] Do not add "Set WallPaper" menu item if not Photo  
 - [x] Some related code style improvement


zeebok (approved):
This code seems reasonable to me!

jeremypw:
@zeebok Thanks for the review! I replaced an `assert` with an `if` clause for release just in case - I trust that is OK.

PR #831, opened 2025-10-31
docs: simplify installation of dev environment

spotlesscoder:


ryonakano:
Sorry, we don't want to write this to keep this distro-agnostic

spotlesscoder:
what about adding distro-specific setup scripts in an extra folder?

spotlesscoder:
awesome, thank you :)

PR #840, opened 2026-05-20
Split rotate left and right buttons

### LINKED ISSUE #745: Feature: Rotate left
### Problem

I want to rotate the image left. I have to click 3 times on rotate right to get it.

### Proposal

Add a button next to rotate right.

### Prior Art (Optional)

Apples Preview app has a rotate left button that turns into rotate right when holding the option key.

Source: https://support.apple.com/guide/preview/resize-rotate-or-flip-an-image-prvw2015/mac

Eye of GNOME (eog) has rotate left and rotate right button.

Screenshot from 2023-07-28 21-10-14

https://wiki.gnome.org/Apps/EyeOfGnome
https://help.gnome.org/users/eog/stable/flip-rotate.html.en

Adobe Lightroom Classic has rotate left and right in the menu with keyboard shortcut.

https://helpx.adobe.com/lightroom-classic/help/retouch-photos.html

stan-janssen:
Fixes #745

This solves feature request #745 by splitting the two rotate actions into separate buttons:

afbeelding

(If this is desirable, I will create a second PR for doing the same to the Flip button).

### DIFF
--- a/src/Resources.vala
+++ b/src/Resources.vala
@@ -78,7 +78,7 @@ public const string ICON_FLAGGED_PAGE = "edit-flag";
 
 public const string ROTATE_CW_MENU = _("Rotate _Right");
 public const string ROTATE_CW_FULL_LABEL = _("Rotate Right");
-public const string ROTATE_CW_TOOLTIP = _("Rotate the photos right (press Ctrl to rotate left)");
+public const string ROTATE_CW_TOOLTIP = _("Rotate the photos right");
 
 public const string ROTATE_CCW_MENU = _("Rotate _Left");
 public const string ROTATE_CCW_FULL_LABEL = _("Rotate Left");
--- a/src/Views/CollectionPage.vala
+++ b/src/Views/CollectionPage.vala
@@ -45,7 +45,8 @@ public abstract class CollectionPage : MediaPage {
     private Gtk.ToggleButton? enhance_button = null;
     private Gtk.Menu item_context_menu;
     private Gtk.Menu contractor_submenu;
-    private Gtk.Button rotate_button;
+    private Gtk.Button rotate_left_button;
+    private Gtk.Button rotate_right_button;
     private Gtk.Button flip_button;
 
     protected CollectionPage (string page_name) {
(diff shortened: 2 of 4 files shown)


ryonakano:
Hi, please do not update the translation files (.pot and .po files) in your commit because that causes your PR bigger and easily can cause conflicts with translation updates from Weblate. We already have a CI that updates them automatically if there are string changes in a pushed commit.

stan-janssen:
How does that work? Can the CI automatically preserve the relevant parts of the existing translations, or does every language lose the translation when a string changes?

ryonakano:
Automatically preserve the relevant parts of the existing translations.

Eventually it just runs `ninja -C`, which should be exactly what you should have done when creating this PR, to update the .pot file. If you're curious about its implementation, it's available at https://github.com/elementary/actions/tree/main/gettext-template.

Then, [Update PO files to match POT (msgmerge)](https://docs.weblate.org/en/weblate-2026.5/admin/addons.html#update-po-files-to-match-pot-msgmerge) Weblate add-on which is installed in [our Weblate instance](https://l10n.elementaryos.org/) updates the .po files in accordance with the .pot file.

stan-janssen:
All right, I'll remove the updates to the translations from the PR.

stan-janssen:
Thanks for pointing me to the Gettext implementation. Just so that I understand: I added changes like this:

```
--- msgid "Rotate the photos right (press Ctrl to rotate left)"
--- msgstr "写真を右に回転します (Ctrl を押すと左に回転します)"
+++ msgid "Rotate the photos right"
+++ msgstr "写真を右に回転します"
```

I can't find how the gettext diff system can automatically figure out what part of the original Japanese string should be kept in this case. Or is this something that Weblate does automatically?

I can see:

> This add-on is essential when working with gettext PO files where the POT (template) file is updated with new strings or changes to existing strings. When the POT file is updated in the repository, this add-on automatically merges those changes into all language PO files, ensuring that:
>
> - New translatable strings appear in all languages
> - Removed strings are marked as obsolete
> - Modified strings are marked as needing review (fuzzy)

I assume that the new string (without the part in parentheses) will be marked as fuzzy and need review, but surely it cannot automatically do what I did and remove the parenthesized part from the translation, right?

Can I preview the automatic changes locally? (`ninja -C` did not work for me).

I don't want to leave the parenthesized instruction there in every language if that feature is no longer present.

ryonakano:
Note that I'm assuming that `I added changes like this:` means you modified the translations to follow change of the source string (I haven't checked the content of translation files in my previous review).

> I can't find how the gettext diff system can automatically figure out what part of the original Japanese string should be kept in this case. Or is this something that Weblate does automatically?

No, this is what translators do. Neither gettext, `ninja -C`, our gettext-template actions, or Weblate add-on can handle.

This is because we can't assume that a corresponding translation of all languages _always_ use parentheses—U+0028 `(` and U+0029 `)`—in the translation for an English string that uses them. For example Japanese or Chinese translators might choose U+FF08 `（` and U+FF09 `）`, aka Zenkaku parentheses. Some translators might choose to translate without using parentheses for better translation. This time it just happens to that case though:

https://l10n.elementaryos.org/translate/photos/-/en/?q=Rotate+the+photos+right+%28press+Ctrl+to+rotate+left%29&sort_by=-priority%2Cposition&checksum=#translations

> I assume that the new string (without the part in parentheses) will be marked as fuzzy and need review, but surely it cannot automatically do what I did and remove the parenthesized part from the translation, right?

Yes, it will look like this in this case:

```po
#: src/Resources.vala:81
#, fuzzy
#| msgid "Rotate the photos right (press Ctrl to rotate left)"
msgid "Rotate the photos right"
msgstr "写真を右に回転します (Ctrl を押すと左に回転します)"
```

The old source string remains after `#|` as a reference for translators, then new source string follows, and finally the old translation that translators should review.

> I don't want to leave the parenthesized instruction there in every language if that feature is no longer present.

If you mean leaving them in the .po files, they should be left for translators to update the translation easily just by modifying the old string.
If you're concerning that they might be shown in the UI, that won't happen. In general the fuzzy marked translations won't be shown in UI and the source string, English (United States) in our case, will be shown instead until translators review the translation and remove the fuzzy flag.

Generally what we elementary developers do is just update the source string in our source code. A .pot file will be updated through CI. Merging changes in .pot file into .po files will be done by Weblate add-on. And finally translators review and update translations in .po files.

stan-janssen:
Thanks for taking the time to explain. I don't quite get what was wrong with my initial PR which already contained a careful edit of all the existing translations so that they don't need to be updated again. I went through all of them, one by one, and removed the parenthesized translation from the translation of the "Rotate the photos right" strings. 

If my branch is up-to-date with the `master` branch just before it gets merged, why would that cause a conflict? Is there a time delay between weblate and this repository?

I don't want people, for any amount of time, to experience a correct translation on the "Rotate image to the left" button (whose translation string did not change), but see the "Rotate image to the right" in English until someone gets around to updating the string in Weblate for their language, and then waiting for a new release to be created that includes the new translation. I imagine that it could take months or year to reach all languages, and for what? 

I don't think this feature is worth that, so I'll close this PR for now. It's not quite a big deal, I just thought I'd fix something that someone asked for and I tried to be complete in my PR so that I don't cause a regression.

Thanks for taking the time and for your help anyway.

PR #848, opened 2026-08-12
Add Send via Email feature

### LINKED ISSUE #624: How to share photos via email with this app.
I do see methods for sharing photos using plug-ins for various online services like Facebook, Flickr, Google and others, none of whom I would ever wish to share one of my photos with, but to each their own I guess.

So, how does a person send a photo to an email recipient from within the app?

zeebok:
Closes #624

### DIFF
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -365,8 +365,11 @@ public abstract class AppWindow : PageWindow {
         try {
             portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
         } catch (Error error) {
-            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
-            message_dialog.transient_for = this;
+            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                _("Could not send selected photos"), error.message, "dialog-error"
+            ) {
+                transient_for = this
+            };
             message_dialog.response.connect (message_dialog.destroy);
             message_dialog.run ();
         }
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -365,7 +365,7 @@ public abstract class AppWindow : PageWindow {
         try {
             portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
         } catch (Error error) {
-            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name ("Could not send selected photos", error.message, "dialog-error");
+            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
             message_dialog.run ();
             message_dialog.destroy ();
         }
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -367,8 +367,8 @@ public abstract class AppWindow : PageWindow {
         } catch (Error error) {
             var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
             message_dialog.transient_for = this;
+            message_dialog.response.connect (message_dialog.destroy);
             message_dialog.run ();
-            message_dialog.destroy ();
         }
     }
 
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -52,6 +52,7 @@ public abstract class AppWindow : PageWindow {
     public const string ACTION_SELECT_NONE = "action_select_none";
     public const string ACTION_UNDO = "action_undo";
     public const string ACTION_SET_WALLPAPER = "set-wallpaper";
+    public const string ACTION_SEND_EMAIL = "send-email";
 
     private const ActionEntry[] ACTION_ENTRIES = {
         { ACTION_FULLSCREEN, on_fullscreen },
@@ -61,7 +62,8 @@ public abstract class AppWindow : PageWindow {
         { ACTION_SELECT_ALL, on_select_all },
         { ACTION_SELECT_NONE, on_select_none },
         { ACTION_UNDO, on_undo },
-        { ACTION_SET_WALLPAPER, action_set_wallpaper, "s" }
+        { ACTION_SET_WALLPAPER, action_set_wallpaper, "s" },
+        { ACTION_SEND_EMAIL, action_send_email, "as" }
     };
 
     protected AppWindow () {
@@ -354,6 +356,13 @@ public abstract class AppWindow : PageWindow {
 
         var portal = new Xdp.Portal ();
         portal.set_wallpaper.begin (parent, param.get_string (), NONE, null);
+    }
+
+    private void action_send_email (GLib.SimpleAction action, GLib.Variant? param) {
+        Xdp.Parent? parent = Xdp.parent_new_gtk (this);
+
+        var portal = new Xdp.Portal ();
+        portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
     }
 
     private void on_undo () {
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -366,6 +366,7 @@ public abstract class AppWindow : PageWindow {
             portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
         } catch (Error error) {
             var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
+            message_dialog.transient_for = this;
             message_dialog.run ();
             message_dialog.destroy ();
         }
--- a/src/AppWindow.vala
+++ b/src/AppWindow.vala
@@ -362,7 +362,13 @@ public abstract class AppWindow : PageWindow {
         Xdp.Parent? parent = Xdp.parent_new_gtk (this);
 
         var portal = new Xdp.Portal ();
-        portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
+        try {
+            portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
+        } catch (Error error) {
+            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name ("Could not send selected photos", error.message, "dialog-error");
+            message_dialog.run ();
+            message_dialog.destroy ();
+        }
     }
 
     private void on_undo () {
--- a/src/Views/CollectionPage.vala
+++ b/src/Views/CollectionPage.vala
@@ -212,17 +212,37 @@ public abstract class CollectionPage : MediaPage {
             contractor_submenu.add (print_menu_item);
             contractor_submenu.add (export_menu_item);
 
-            var source = get_view ().get_selected_at (0).source;
-            if (source != null && source is Photo) {
-                var file = ((Photo) source).get_file ();
-                if (file != null) {
-                    var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
-                        action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER,
-                        action_target = new Variant.string (file.get_uri ())
-                    };
-                    contractor_submenu.add (wallpaper_menuitem);
+            var sources = get_view ().get_selected_sources ();
+
+            if (sources.size == 1) {
+                var source = sources.get(0);
+                if (source != null && source is Photo) {
+                    var file = ((Photo) source).get_file ();
+                    if (file != null) {
+                        var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
+                            action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER,
+                            action_target = new Variant.string (file.get_uri ())
+                        };
+                        contractor_submenu.add (wallpaper_menuitem);
+                    }
                 }
             }
+
+            string[] source_uris = new string[sources.size];
+            for (int i = 0; i < sources.size; i++) {
+                var source = sources.get (i);
+                if (source != null && source is Photo) {
+                    var file = ((Photo) source).get_file ();
+                    if (file != null) {
+                        source_uris[i] = file.get_path ();
+                    }
+                }
+            }
+            var email_menuitem = new Gtk.MenuItem.with_label (_("Send via Email")) {
+                action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SEND_EMAIL,
+                action_target = new Variant.strv (source_uris)
+            };
+            contractor_submenu.add (email_menuitem);
 
             item_context_menu.add (copy_images_menu_item);
             item_context_menu.add (adjust_datetime_menu_item);
--- a/src/Views/CollectionPage.vala
+++ b/src/Views/CollectionPage.vala
@@ -215,7 +215,7 @@ public abstract class CollectionPage : MediaPage {
             var sources = get_view ().get_selected_sources ();
 
             if (sources.size == 1) {
-                var source = sources.get(0);
+                var source = sources.get (0);
                 if (source != null && source is Photo) {
                     var file = ((Photo) source).get_file ();
                     if (file != null) {
--- a/src/Views/CollectionPage.vala
+++ b/src/Views/CollectionPage.vala
@@ -211,37 +211,7 @@ public abstract class CollectionPage : MediaPage {
             contractor_menu_item.set_submenu (contractor_submenu);
             contractor_submenu.add (print_menu_item);
             contractor_submenu.add (export_menu_item);
-
-            var sources = get_view ().get_selected_sources ();
-
-            if (sources.size == 1) {
-                var source = sources.get (0);
-                if (source != null && source is Photo) {
-                    var file = ((Photo) source).get_file ();
-                    if (file != null) {
-                        var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
-                            action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER,
-                            action_target = new Variant.string (file.get_uri ())
-                        };
-                        contractor_submenu.add (wallpaper_menuitem);
-                    }
-                }
-            }
-
-            string[] source_uris = new string[sources.size];
-            for (int i = 0; i < sources.size; i++) {
-                var source = sources.get (i);
-                if (source != null && source is Photo) {
-                    var file = ((Photo) source).get_file ();
-                    if (file != null) {
-                        source_uris[i] = file.get_path ();
-                    }
-                }
-            }
-            var email_menuitem = new Gtk.MenuItem.with_label (_("Send via Email")) {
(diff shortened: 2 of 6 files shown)


src/Views/CollectionPage.vala
```diff
@@ -212,18 +212,38 @@ public abstract class CollectionPage : MediaPage {
...
-                if (file != null) {
-                    var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
-                        action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER,
-                        action_target = new Variant.string (file.get_uri ())
-                    };
-                    contractor_submenu.add (wallpaper_menuitem);
+            var sources = get_view ().get_selected_sources ();
+
+            if (sources.size == 1) {
+                var source = sources.get(0);
```
ryonakano:
```suggestion
                var source = sources.get (0);
```

src/Views/LibraryPhotoPage.vala
```diff
@@ -567,17 +567,39 @@ public class LibraryPhotoPage : EditingHostPage {
...
-                    var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
-                        action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER,
-                        action_target = new Variant.string (file.get_uri ())
-                    };
-                    contractor_submenu.add (wallpaper_menuitem);
+
+            var sources = get_view ().get_selected_sources ();
+
+            if (sources.size == 1) {
+                var source = sources.get(0);
```
ryonakano:
```suggestion
                var source = sources.get (0);
```

src/Views/CollectionPage.vala
```diff
@@ -212,18 +212,38 @@ public abstract class CollectionPage : MediaPage {
...
+            string[] source_uris = new string[sources.size];
+            for (int i = 0; i < sources.size; i++) {
+                var source = sources.get (i);
+                if (source != null && source is Photo) {
+                    var file = ((Photo) source).get_file ();
+                    if (file != null) {
+                        source_uris[i] = file.get_path ();
+                    }
+                }
+            }
```
ryonakano, line 240: I'm not sure if it's a best solution to use `string[]` here, although `Xdp.Portal.compose_email ()` surely accepts that type via the `attachments` parameter. Reasons:

- Could this break the stack memory and causes SIGSEGV if `sources.size` stores a very large number
  -  I'm not sure how Vala handles `string[]` though; if it's allocated in the heap this would not occur easily
- This wouldn't occur, but `source_uris[]` may contain uninitialized elements if one of the following condition is met. Dose `Xdp.Portal.compose_email ()` work as expected even if it get such parameter?
  - `source` is null
  - `source` is not typeof `Photo`
  - `file` is null

So I wonder if we should use other types like an array or list, and then convert it to an `string[]` in `action_send_email ()`.

(BTW it's unfortunate that Photos is still in GTK 3 and thus `Gtk.StringList` isn't available……)

src/AppWindow.vala
```diff
@@ -356,6 +358,13 @@ public abstract class AppWindow : PageWindow {
         portal.set_wallpaper.begin (parent, param.get_string (), NONE, null);
     }
 
+    private void action_send_email (GLib.SimpleAction action, GLib.Variant? param) {
+        Xdp.Parent? parent = Xdp.parent_new_gtk (this);
+
+        var portal = new Xdp.Portal ();
+        portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
```
ryonakano: We should handle Error of `Xdp.Portal.compose_email ()`.

src/Views/CollectionPage.vala
zeebok: I initially misread and was passing the URI instead of the path and `compose_email` failed gracefully. It printed output and simply did not add anything as an attachment. I imagine it handles null strings in the same way.

I agree with you that `string[]` feels bad but I am not sure what safer container there is that can also easily go back to `string[]`. I could force an upper limit to the number of attachments or figure out how to have a total size limit.

src/AppWindow.vala
zeebok: Unless there is a better idea I will simply make a MessageDialog stating there was an error and the error message as the body.

src/AppWindow.vala
```diff
@@ -356,6 +358,19 @@ public abstract class AppWindow : PageWindow {
...
     }
 
+    private void action_send_email (GLib.SimpleAction action, GLib.Variant? param) {
+        Xdp.Parent? parent = Xdp.parent_new_gtk (this);
+
+        var portal = new Xdp.Portal ();
+        try {
+            portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
+        } catch (Error error) {
+            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
```
ryonakano, line 368:
```suggestion
            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
            message_dialog.transient_for = this;
```

This dims the background of the dialog, which allows us to get attention of the user.

src/AppWindow.vala
```diff
@@ -356,6 +358,19 @@ public abstract class AppWindow : PageWindow {
...
+            message_dialog.run ();
+            message_dialog.destroy ();
```
ryonakano:
```suggestion
            message_dialog.response.connect (message_dialog.destroy);
            message_dialog.run ();
```

GTK 4 preparation. We would need to replace `message_dialog.run ()` with `message_dialog.present ()` when porting to GTK 4 but IIRC `Gtk.Window.present ()` isn't a blocking operation, resulting `message_dialog.destroy ()` is called soon after `message_dialog.present ()`. So we need to connect to `response` signal in GTK 4 beforehand.

But this is an error route and we can potentially forget to do this, so I think we should consider this now so that we just only need to replace `message_dialog.run ()` with `message_dialog.present ()` when porting to GTK 4.

ryonakano (commented):
I'm not sure if these are related to this PR, but I noticed that the content of `sources` possibly cached.

- "Set as Wallpaper" always appears even if you select how many images if you select one image and then open the context menu beforehand
- The images that can be attached using "Send via Email" is not changed if you change selected images

src/Views/CollectionPage.vala
ryonakano: I looked at the C code that valac generated and `string[]` is represented by `g_new0()` which means it's allocated on the heap memory. So I guess we don't need to consider my first concern.

```c
// builddir/src/io.elementary.photos.p/Views/CollectionPage.c

#line 231 "../src/Views/CollectionPage.vala"
                _tmp133_ = _tmp132_;
#line 231 "../src/Views/CollectionPage.vala"
                _tmp134_ = g_new0 (gchar*, _tmp133_ + 1);
#line 231 "../src/Views/CollectionPage.vala"
                source_uris = _tmp134_;
#line 231 "../src/Views/CollectionPage.vala"
                source_uris_length1 = _tmp133_;
#line 231 "../src/Views/CollectionPage.vala"
                _source_uris_size_ = source_uris_length1;
```

ryonakano: @zeebok I think we should consider my second concern. Photos will crash if you try to send non-image files like videos via email:

```
user@elementary-8-daily:~/work/photos$ io.elementary.photos 
** Message: 07:30:51.165: main.vala:301: Photos Photo Manager 8.0.2
** Message: 07:30:51.179: main.vala:52: Verifying database …

(io.elementary.photos:2503): GLib-CRITICAL **: 07:30:57.116: g_variant_new_string: assertion 'string != NULL' failed

(io.elementary.photos:2503): GLib-CRITICAL **: 07:30:57.116: g_variant_ref_sink: assertion 'value != NULL' failed
Segmentation fault (core dumped)
user@elementary-8-daily:~/work/photos$ 
```

src/AppWindow.vala
zeebok: That works for me. I am also currently in the process of rewriting Photos from scratch in GTK 4 so I will make sure to note this for that branch as well!

src/Views/CollectionPage.vala
zeebok: Hmm alright surprised it crashes since it should just skip the file if it can't be properly cast to a Photo object. I will take a look though

zeebok: > I looked at the C code that valac generated and `string[]` is represented by `g_new0()` which means it's allocated on the heap memory. So I guess we don't need to consider my first concern.

That's good news but emails usually have a 10mb limit by default so I might prevent linking the action target if there is too large of a collective size on the selected images.

zeebok:
> I'm not sure if these are related to this PR, but I noticed that the content of `sources` possibly cached.
> 
>     * "Set as Wallpaper" always appears even if you select how many images if you select one image and then open the context menu beforehand
> 
>     * The images that can be attached using "Send via Email" is not changed if you change selected images

Ah I see, the menu itself gets cached it seems. I will fix this when I have a moment!

src/Views/Page.vala
```diff
@@ -100,6 +112,44 @@ public abstract class Page : Gtk.ScrolledWindow {
...
+        if (sources.size == 1) {
+            var source = sources.get (0);
+            if (source != null) {
+                var file = ((Photo) source).get_file ();
+                if (file != null) {
+                        wallpaper_menuitem.action_target = new Variant.string (file.get_uri ());
+                }
+            }
+        }
+        else {
```
ryonakano:
```suggestion
        } else {
```

Our coding style doesn't have a newline before `else`

https://docs.elementary.io/develop/writing-apps/code-style

src/Views/Page.vala
```diff
@@ -100,6 +112,44 @@ public abstract class Page : Gtk.ScrolledWindow {
...
 
+    protected void populate_portal_menu_items (Gtk.Menu menu) {
+        var sources = get_view ().get_selected_sources_of_type (typeof (Photo));
+
+        if (sources.size == 1) {
+            var source = sources.get (0);
+            if (source != null) {
+                var file = ((Photo) source).get_file ();
+                if (file != null) {
+                        wallpaper_menuitem.action_target = new Variant.string (file.get_uri ());
```
ryonakano:
```suggestion
                    wallpaper_menuitem.action_target = new Variant.string (file.get_uri ());
```

Strange indentation

src/AppWindow.vala
```diff
@@ -356,6 +358,20 @@ public abstract class AppWindow : PageWindow {
...
 
+    private void action_send_email (GLib.SimpleAction action, GLib.Variant? param) {
+        Xdp.Parent? parent = Xdp.parent_new_gtk (this);
+
+        var portal = new Xdp.Portal ();
+        try {
+            portal.compose_email (parent, null, null, null, null, null, param.get_strv (), NONE, null);
+        } catch (Error error) {
+            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (_("Could not send selected photos"), error.message, "dialog-error");
+            message_dialog.transient_for = this;
```
ryonakano:
```suggestion
            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
                                    _("Could not send selected photos"), error.message, "dialog-error"
            ) {
                transient_for = this
            };
```

The first line is too long and plus we can use GObject style construction here.

src/Views/Page.vala
```diff
@@ -19,6 +19,7 @@
 
 public abstract class Page : Gtk.ScrolledWindow {
     private const int CONSIDER_CONFIGURE_HALTED_MSEC = 400;
+    private const int64 MAX_EMAIL_ATTACH_SIZE = 10485760; // 10MB in bytes, typical email limit
```
ryonakano:
```vala
    private const int64 MAX_EMAIL_ATTACH_BYTES = 10 * 1024 * 1024;
```

We should use an expression instead of comments which is more accurate. Also explicitly name that this is in bytes.

ryonakano (commented):
Sorry for my late review.

zeebok:
> Sorry for my late review.

@ryonakano Thanks for reviewing! The code style at work is bleeding into my non-job code, I hate it :sweat_smile:

src/Views/DirectPhotoPage.vala
```diff
@@ -164,14 +164,9 @@ public class DirectPhotoPage : EditingHostPage {
...
                 print_menu_item.activate.connect (() => print_action.activate ());
 
-                var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
-                    action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER
-                };
-
                 var contractor_menu_item = new Gtk.MenuItem.with_mnemonic (_("Other Actions"));
                 contractor_menu = new Gtk.Menu ();
                 contractor_menu.add (print_menu_item);
-                contractor_menu.add (wallpaper_menuitem);
```
ryonakano, line 174: Now that you can't set a currently displaying picture as wallpaper, which I don't think expected change in this PR.

src/Views/Page.vala
```diff
@@ -100,6 +112,43 @@ public abstract class Page : Gtk.ScrolledWindow {
...
+        if (sources.size == 1) {
+            var source = sources.get (0);
+            if (source != null) {
+                var file = ((Photo) source).get_file ();
+                if (file != null) {
+                    wallpaper_menuitem.action_target = new Variant.string (file.get_uri ());
+                }
+            }
+        } else {
+            wallpaper_menuitem.action_target = null;
```
ryonakano: This seems to cause the following warning when selecting multiple files:

```
(io.elementary.photos:11449): Gtk-WARNING **: 15:56:01.162: actionhelper: action win.set-wallpaper can't be activated due to parameter type mismatch (parameter type s, target type NULL)
```

Same comment goes to `email_menuitem.action_target = null;` too.

src/Views/Page.vala
```diff
@@ -100,6 +112,43 @@ public abstract class Page : Gtk.ScrolledWindow {
...
+                    source_uris[i] = file.get_path ();
+                }
+            }
+        }
+        if (source_uris.length > 0 && total_size < MAX_EMAIL_ATTACH_SIZE) {
+            email_menuitem.action_target = new Variant.strv (source_uris);
+        }
+        else {
+            email_menuitem.action_target = null;
+        }
```
ryonakano, line 148:
```suggestion
        }

        if (source_uris.length > 0 && total_size < MAX_EMAIL_ATTACH_SIZE) {
            email_menuitem.action_target = new Variant.strv (source_uris);
        } else {
            email_menuitem.action_target = null;
        }
```

Another if-else coding style error and plus missing newline after a closing bracket

ryonakano (commented):
Sorry for my late review again.

src/Views/Page.vala
zeebok: D'oh!

src/Views/DirectPhotoPage.vala
zeebok: It is part of the populate_portal_menu_items call. It is technically unnecessary since caching the single photo of the page won't cause issues but I also wanted to follow DRY here.

zeebok: Or at least it *was* getting populated by that. For some reason now it doesn't populate it with either portal menu item.

zeebok: Fixed it. Somehow the add calls were removed before pushing? The menu items should appear in all views of photos now.

src/Views/Page.vala
zeebok: Ah the call to set the target is what I want to use!

src/Views/Page.vala
```diff
@@ -100,6 +112,47 @@ public abstract class Page : Gtk.ScrolledWindow {
...
+                var file_info = file.query_info ("standard::size", FileQueryInfoFlags.NONE);
+                total_size += file_info.get_size ();
+
+                if (file != null) {
+                    source_uris[i] = file.get_path ();
+                }
+            }
+        }
+        if (source_uris.length > 0 && total_size < MAX_EMAIL_ATTACH_SIZE) {
+            email_menuitem.action_target = new Variant.strv (source_uris);
```
ryonakano, line 145:
```suggestion
        }

        if (source_uris.length > 0 && total_size < MAX_EMAIL_ATTACH_SIZE) {
```

Missing newline after closing brackets.

src/Views/Page.vala
```diff
@@ -100,6 +112,47 @@ public abstract class Page : Gtk.ScrolledWindow {
...
 
+    protected void populate_portal_menu_items (Gtk.Menu menu) {
+        var sources = get_view ().get_selected_sources_of_type (typeof (Photo));
+
+        if (sources.size == 1) {
+            var source = sources.get (0);
+            if (source != null) {
+                var file = ((Photo) source).get_file ();
+                if (file != null) {
+                    wallpaper_menuitem.action_target = new Variant.string (file.get_uri ());
```
ryonakano: I think we should unify to `set_action_target_value ()` for consistency; same comment applies to all occurrences of `action_target` in this PR.

src/Views/Page.vala
```diff
@@ -100,6 +112,47 @@ public abstract class Page : Gtk.ScrolledWindow {
...
+                }
+            }
+        } else {
+            wallpaper_menuitem.set_action_target_value (null);
```
ryonakano, line 127: I'm still seeing the following warnings:

```
(io.elementary.photos:7399): Gtk-WARNING **: 21:56:03.363: actionhelper: action win.set-wallpaper can't be activated due to parameter type mismatch (parameter type s, target type NULL)

(io.elementary.photos:7399): Gtk-WARNING **: 21:56:03.366: actionhelper: action win.send-email can't be activated due to parameter type mismatch (parameter type as, target type NULL)
```

src/Views/CollectionPage.vala
```diff
@@ -211,18 +211,8 @@ public abstract class CollectionPage : MediaPage {
...
-                if (file != null) {
-                    var wallpaper_menuitem = new Gtk.MenuItem.with_label (_("Set as Wallpaper")) {
-                        action_name = AppWindow.ACTION_PREFIX + AppWindow.ACTION_SET_WALLPAPER,
-                        action_target = new Variant.string (file.get_uri ())
-                    };
-                    contractor_submenu.add (wallpaper_menuitem);
-                }
-            }
+            contractor_submenu.add (wallpaper_menuitem);
+            contractor_submenu.add (email_menuitem);
```
ryonakano, line 215: I'm seeing the following warnings:

```
(io.elementary.photos:7399): Gtk-WARNING **: 21:56:03.366: Attempting to add a widget with type GtkMenuItem to a container of type GtkMenu, but the widget is already inside a container of type GtkMenu, please remove the widget from its existing container first.

(io.elementary.photos:7399): Gtk-WARNING **: 21:56:03.366: Attempting to add a widget with type GtkMenuItem to a container of type GtkMenu, but the widget is already inside a container of type GtkMenu, please remove the widget from its existing container first.
```

src/Views/Page.vala
zeebok: Not sure what do about this, the [documentation here](https://valadoc.org/gtk4/Gtk.Actionable.set_action_target_value.html) says that to unset the action you pass it null.

zeebok: Oh I think I found a solution

